PR #34, opened 2020-06-24
Add codecov.io action

fingon:


fingon:
Doesn't seem to work quite. I guess I'll make just normal htmlcov target for now.

PR #35, opened 2020-08-24
Add metrics

tsharju:
Collects the following metrics from running operations:

  * `astacus_fail`, a counter reporting all failed operations, tagged with the op type
  * `astacus_step_duration`, a timing stat for each step of an operation, tagged with op type and step name
  * `astacus_lock_call_failure`,  a counter reporting locking failures, tagged with call (lock/unlock) and locker
  * `astacus_op_running_for`, a gauge with seconds the op has run as value, tagged with op type and id

### DIFF
--- a/astacus/coordinator/coordinator.py
+++ b/astacus/coordinator/coordinator.py
@@ -147,11 +147,6 @@ class CoordinatorOp(op.Op):
             logger.info("%s - permanent failure: %r", name, ex)
         self.set_status_fail()
 
-    def set_status_fail(self):
-        super().set_status_fail()
-        name = self.__class__.__name__
-        self.stats.increase("astacus_fail", {"op": name})
-
     async def wait_successful_results(self, start_results, *, result_class, all_nodes=True):
         urls = []
         for i, result in enumerate(start_results, 1):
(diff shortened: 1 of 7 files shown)


astacus/coordinator/coordinator.py
```diff
@@ -239,6 +239,20 @@ async def run(self):
...
+        changed = super().set_status(state=state, from_state=from_state)
+        self._update_running_stats()
+        return changed
+
+    def _update_running_stats(self):
+        if self.stats is not None:
+            if self.info.op_status in {op.Op.Status.done, op.Op.Status.fail}:
+                op_running = 0
+            else:
+                op_running = 1
```
fingon, line 252: I'd rather have op_running_for metric, which would have number that is ~
`int(time.monotonic()-start_time_time_monotonic+1)`.

With this, we can create alerting which is e.g. `astacus_op_running_for > 3600` to detect ops that take over hour without any hard lifting on the alerting side.

( Disclaimer: If this case is somehow covered elsewhere that's fine too, I just didn't notice it )

fingon (changes requested):

astacus/coordinator/coordinator.py
tsharju: True. I initially had that, but then went for the "binary" approach with just 1 or 0. I'll fix this.

fingon (approved):

PR #38, opened 2020-10-05
exception handling fixes

fingon:
The previous version was broken (CancelledError was caught in the blanket exception handling). Attempt to fix that + few things that spam sentry needlessly.

### DIFF
--- a/astacus/common/op.py
+++ b/astacus/common/op.py
@@ -20,6 +20,7 @@ from fastapi import HTTPException
 from typing import Optional
 from urllib.parse import urlunsplit
 
+import asyncio
 import contextlib
 import inspect
 import logging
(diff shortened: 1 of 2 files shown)


astacus/common/op.py
```diff
@@ -98,6 +98,8 @@ async def _async_wrapper():
                 op.set_status(Op.Status.running)
                 await fun()
                 op.set_status(Op.Status.done, from_state=Op.Status.running)
+            except ExpiredOperationException:
+                pass
```
tsharju, line 103: Maybe wrap the whole wrapper with `contextlib.suppress(ExpiredOperationException)`?

astacus/common/op.py
```diff
@@ -109,6 +111,8 @@ def _sync_wrapper():
                 op.set_status(Op.Status.running)
                 fun()
                 op.set_status(Op.Status.done, from_state=Op.Status.running)
+            except ExpiredOperationException:
+                pass
```
tsharju, line 119: Same as above.

tsharju (dismissed):

astacus/common/op.py
fingon: Unfortunately that isn't possible, as it isn't async context manager (with just plain 'with ..', it would get ignored also in other async threads for duration of this async op).

astacus/common/op.py
fingon: Done (sync case is now more elegant).

tsharju (approved):

PR #42, opened 2020-12-10
coordinator: Restore downloads snapshots from requested storage

ilpon:


tests/unit/coordinator/test_restore.py
```diff
@@ -41,9 +42,11 @@
 )
 
 
+@pytest.mark.parametrize("storage_name", [None, "x", "y"])
```
fingon, line 45: This leads to 3 * 4 = 12 tests. I don't think all of the combinations are really useful?

fingon (dismissed):
LGTM, test could use some refactoring though as adding 9 more tests due to this feels excessive and they are pretty duplicate ones too.

fingon (approved):

PR #55, opened 2021-06-16
WIP Cassandra support

fingon:
Leaving this draft here mostly to track how it is going in terms of CI here. Will be probably reopened once it is close to being ready.


fingon:
New replacement PR is up, leaving the old branch around until the replacement code is actually done.

PR #58, opened 2021-09-28
misc refactoring

kmichel-aiven:
This is mostly preparation work to reduce coupling, make it easier to test each part
separately and add more plugins (next one being ClickHouse: #57).

Addition of type hints almost everywhere led mypy to expose a few typing inconsistencies
that explain some of the code movements.

I plan to add more tests for the m3db part, but this PR is already quite large.


dmitry-potepalov (commented):
commits about extracting cluster functions and backup/restore split I'd like to discuss, a little bit too big to digest.
And this PR requires further review, I'm not familiar enough with astacus to pronounce a final verdict :).

astacus/common/utils.py
```diff
@@ -109,12 +109,12 @@ async def httpx_request(url, *, caller, method="get", timeout=10, json: bool = T
...
                 return r.json() if json else r
             logger.warning("Unexpected response status code from %s to %s: %s %r", url, caller, r.status_code, r.text)
-        except httpcore.ConnectError:
+        except httpcore.NetworkError as ex:
             # Unfortunately at least current httpx leaks this
             # exception without wrapping it. Future versions may
             # address this hopefully. I believe httpx.TransportError
             # replaces it in future versions once we upgrade.
-            pass
+            logger.warning("Network error from %s to %s: %r", url, caller, ex)
```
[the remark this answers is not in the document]
kmichel-aiven: Yes, NetworkError is a parent of ConnectError but still part of the wrapped httpx.TransportError in future versions.

astacus/coordinator/plugins/m3db.py
```diff
@@ -133,7 +133,7 @@ async def step_restore_etcd(self):
         if self.req.partial_restore_nodes:
             logger.debug("Skipping etcd restoration due to partial backup restoration")
             return True
-        return await self.restore_etcd_dump(self.result_rewrite_etcd)
+        return await restore_etcd_dump(ETCDClient(self.plugin_config.etcd_url), self.result_rewrite_etcd)
```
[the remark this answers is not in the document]
kmichel-aiven: This is replaced in later commits :)

astacus/common/utils.py
```diff
@@ -128,7 +128,22 @@ async def httpx_request(
...
+
+
+def exponential_backoff(
+    *,
+    initial: float,
+    retries: Optional[int] = None,
+    multiplier: float = 2,
+    maximum: Optional[float] = None,
+    duration: Optional[float] = None,
+    wakeup_event: Optional[asyncio.Event] = None
```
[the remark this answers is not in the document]
kmichel-aiven: I've encapsulated the entire sleep/wakeup logic, it looks better :+1:

astacus/coordinator/plugins/files.py
```diff
@@ -15,6 +15,7 @@
 
 """
 
+from ..cluster import Cluster
```
fingon: Please use absolute path in non-local imports (this is in few other places too, but just noting it where I first noticed it)

astacus/coordinator/coordinator.py
```diff
@@ -221,31 +246,82 @@ def get_subresult_url(request_url: URL, op_id: int) -> str:
     return urlunsplit(parts)
 
 
-class Coordinator(op.OpMixin):
-    """ Convenience dependency which contains sub-dependencies most API endpoints need """
-    state: CoordinatorState
+class BackupRestoreOp(CoordinatorOpWithClusterLock):
```
fingon, line 249: This name is pretty bad - it can be also construed as backup restoration (although it is really utility superclass). Could you figure something else for it? As it seems to mostly deal with the steps, perhaps StepOp(Base) or something.

astacus/coordinator/coordinator.py
```diff
@@ -221,31 +246,82 @@ def get_subresult_url(request_url: URL, op_id: int) -> str:
...
+                with self._progress_handler(cluster, step):
+                    r = await step.run_step(cluster, context)
+            if not r:
+                logger.info("Step %s failed", step)
+                return False
+            context.set_result(step.__class__, r)
+        return True
+
+    def _timing_manager(self, metric: str, tags: Optional[Tags] = None) -> AsyncContextManager:
+        if self.stats is None:
```
fingon, line 299: This seems unnecessary, as the stats probably should be provided in the superclasses? (If not, that should be fixed - non-stats case is aberration we don't really want)

astacus/coordinator/coordinator.py
```diff
@@ -221,31 +246,82 @@ def get_subresult_url(request_url: URL, op_id: int) -> str:
...
+                logger.info("Step %s not even started due to shutdown", step_name)
+                return False
+            logger.debug("Step %d/%d: %s", i, len(self.steps), step_name)
+            async with self._timing_manager("astacus_step_duration", {"op": op_name, "step": step_name}):
+                with self._progress_handler(cluster, step):
+                    r = await step.run_step(cluster, context)
+            if not r:
+                logger.info("Step %s failed", step)
+                return False
+            context.set_result(step.__class__, r)
```
fingon, line 295: I think you should also ensure that they key is already not set, just in case.

astacus/coordinator/plugins/base.py
```diff
@@ -5,347 +5,349 @@
...
 
-    def _snapshot_results_to_upload_node_index_datas(self) -> List[NodeIndexData]:
-        assert len(self.result_snapshot) == len(self.nodes)
-        sshash_to_node_indexes: Dict[ipc.SnapshotHash, List[int]] = {}
-        for i, snapshot_result in enumerate(self.result_snapshot):
-            for sshash in snapshot_result.hashes or []:
-                sshash_to_node_indexes.setdefault(sshash, []).append(i)
+    async def run_step(self, cluster: Cluster, context: StepsContext) -> List[ipc.SnapshotResult]:
+        """ Snapshot step. Has to be parametrized with the root_globs to use """
+        logger.debug("SnapshotStep")
```
fingon, line 72: Leftover?

fingon (dismissed):
Overall, great work. Some stylistic comments. Have you tested that the M3 parts actually work? (The tests here are pretty limited for that, compared to e.g. systest-long in you-know-where)

As aside, once the code stabilizes (= no big things in the pipe) we should move to Black from yapf. But let's not do it before Cassandra and Clickhouse are in the bag :-) 

(Should probably stick that to the TODO)

astacus/coordinator/coordinator.py
kmichel-aiven: Fixed in the StepsContext.

astacus/coordinator/coordinator.py
kmichel-aiven: Yes, it's actually better to have stats non-nullable, if someone really wants to disable stats, a stats object without config just does nothing.

fingon (approved):
LGTM (some sort of factory method for creating op class instances from coordinator might have been nice, as now `op_id` and `stats` passing seems to duplicate code quite a bit, but I guess this works too). Only question is, does it actually work still with the existing M3 code :)

kmichel-aiven:
> LGTM (some sort of factory method for creating op class instances from coordinator might have been nice, as now `op_id` and `stats` passing seems to duplicate code quite a bit, but I guess this works too). Only question is, does it actually work still with the existing M3 code :)

It works with M3 :) (long system tests passed and I also tried a few manual backup/restore).

For the factory that's a good point, but I'm wondering if it could be better to rely less on the Coordinator and more on the FastAPI dependency injection to allocate the op_id and provide stats (and maybe even the storages).

fingon:
> For the factory that's a good point, but I'm wondering if it could be better to rely less on the Coordinator and more on the FastAPI dependency injection to allocate the op_id and provide stats (and maybe even the storages).

Dependency injection would be another solution too. I see just lots of non-DRY in the new code now. Still, merging this one, you can come back to it later if you feel like it.

PR #60, opened 2022-01-02
clickhouse: support restoring Replicated database settings [BF-526]

kmichel-aiven:


carobme (approved):
Don't really know astacus that well, but LGTM.

PR #61, opened 2022-01-22
clickhouse: avoid overloading server during ATTACH and SYNC

kmichel-aiven:
Increase timeout and add a concurrency limiter to avoid
issuing too many queries at once during ATTACH and SYNC
operations on large backups with many file


dmitry-potepalov (approved):
LGTM

PR #62, opened 2022-02-02
clickhouse: support non-utf-8 string

kmichel-aiven:
ClickHouse strings are byte strings and can contain any byte sequence,
including ones that cannot be decoded as UTF-8 or serialized in JSON.

This is true for database names, table names, access entities
and all the queries necessary to create these database objects.


carobme (approved):
LGTM, but I won't pretend to have reviewed the changes very thoroughly. I didn't like the move from `AstacusModel` to `Dict` at first, but the explanation kinda makes sense.

And bonus points for `z_\\x80_enjoyer`, though I sadly never touched one (which probably isn't a big loss as the keyboard is said to be crappy).

PR #63, opened 2022-02-04
docs: improve development instructions

juha-aiven:
Alternative approaches exist, but at least this one should work. The alternatives can be added later.


fingon (changes requested):
LGTM, but setup.py (and preceding `make` which is not even documented) should not be dropped either.

fingon:
Closing due to inaction - I added somewhat better instructions to https://github.com/aiven/astacus/pull/67

PR #68, opened 2022-03-04
common: add azure service principal auth support

raphlopez:
Previous config only allowed for authentication via storage_account name and key, which provides access to all resources in a storage_account. This PR adjusts the schema validation for the config to allow for authentication via service principal (see [Azure docs](https://docs.microsoft.com/en-us/azure/active-directory/fundamentals/service-accounts-principal)). This approach provides granular permissions to be set at the container/blob levels. The service principal authentication itself is [being implemented in pghoard](https://github.com/aiven/pghoard/pull/509), which uses the config to build the AzureTransfer client to perform/restore backups.

### DIFF
--- a/astacus/common/rohmustorage.py
+++ b/astacus/common/rohmustorage.py
@@ -82,11 +82,22 @@ class RohmuS3StorageConfig(RohmuModel):
     # Some more obscure options with defaults are omitted
 
 
+class RohmuAzureServicePrincipalCredentials(RohmuModel):
+    client_id: str
+    secret: str
+
+
+class RohmuAzureServicePrincipalConfig(RohmuModel):
+    client_credentials: RohmuAzureServicePrincipalCredentials
+    tenant_id: str
+
+
 class RohmuAzureStorageConfig(RohmuModel):
     storage_type: Literal[RohmuStorageType.azure]
     account_name: str
-    account_key: str
     bucket_name: str
+    account_key: Optional[str] = None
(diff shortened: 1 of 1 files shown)


astacus/common/rohmustorage.py
```diff
@@ -85,8 +85,9 @@ class RohmuS3StorageConfig(RohmuModel):
 class RohmuAzureStorageConfig(RohmuModel):
     storage_type: Literal[RohmuStorageType.azure]
     account_name: str
-    account_key: str
     bucket_name: str
+    account_key: Optional[str] = None
+    client_config: Optional[Dict[str, Any]] = None
```
fingon, line 90: Please add explicit typing for this too, and also validator which ensures that one or the other is configured. It doesn't seem like client_config is arbitrary dict but instead something with regular structure.

fingon (changes requested):

astacus/common/rohmustorage.py
raphlopez: Explicit typing here certainly makes sense, thanks for pointing that out.
The validation on the other hand is being implemented in pghoard's AzureTransfer client ([PR #509](https://github.com/aiven/pghoard/pull/509)), which astacus instantiates via the `get_transfer()` method. An `InvalidConfigurationError` is thrown if neither `account_key` nor `client_config` is specified.

fingon: Yeah, we just rely on our own configuration format being well-formed (instead of funny exception that shows up later).

fingon (approved):

PR #69, opened 2022-03-11
common: use sas_token for azure auth

raphlopez:
Add sas_token to credential schema as a replacement for Azure Service Principal credentials. The benefits of authenticating using a [SAS token](https://docs.microsoft.com/en-us/azure/storage/common/storage-sas-overview) are two-fold:

1. Simpler credential schema: `sas_token` string instead of `client_id`, `client_secret`, and `tenant_id` required for Service Principal authentication
2. Increased security (better access management): If a SAS token is compromised for any reason, the Shared Access Policy (SAP) it was created with can be deleted to immediately revoke access to the container. On the other hand, if Service Principal credentials are compromised, its credentials must be rotated and redistributed to any service that relied on the original credentials to continue performing backup operations.

### DIFF
--- a/astacus/common/rohmustorage.py
+++ b/astacus/common/rohmustorage.py
@@ -84,22 +84,12 @@ class RohmuS3StorageConfig(RohmuModel):
     # Some more obscure options with defaults are omitted
 
 
-class RohmuAzureServicePrincipalCredentials(RohmuModel):
-    client_id: str
-    secret: str
-
-
-class RohmuAzureServicePrincipalConfig(RohmuModel):
-    client_credentials: RohmuAzureServicePrincipalCredentials
-    tenant_id: str
-
-
 class RohmuAzureStorageConfig(RohmuModel):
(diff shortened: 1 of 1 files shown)


kmichel-aiven:
Just to check:
- this can't be merged until https://github.com/aiven/pghoard/pull/509 is merged first ?
- `client_config` was never supported by pghoard, if was in a previous version of https://github.com/aiven/pghoard/pull/509 but was replaced with `sas_token` even before being merged ?

raphlopez:
Both are correct. The `client_config` approach was never merged in pghoard since the `sas_token` authentication was implemented instead.

fingon (dismissed):
Marking changes requested as it really waits for https://github.com/aiven/rohmu/pull/46 and https://github.com/aiven/astacus/pull/73 to land first. Please dismiss review when both have landed :) (and the rohmu change is in release which is in pypi)

kmichel-aiven (approved):

PR #70, opened 2022-03-13
tests: use ZooKeeper from upstream for tests

kmichel-aiven:
This removes the dependency on docker that is problematic when already
inside a container and the duplication of ZooKeeper fixtures.


carobme (approved):
LGTM

PR #72, opened 2022-04-08
clickhouse: use transaction when creating Replicated entities

kmichel-aiven:
This avoids partial entities that can disturb normal ClickHouse operations
and partially overwriting existing entities.

### DIFF
--- a/astacus/coordinator/plugins/zookeeper.py
+++ b/astacus/coordinator/plugins/zookeeper.py
@@ -81,6 +81,22 @@ class ZooKeeperConnection:
         """
         raise NotImplementedError
 
+    async def try_create(self, path: str, value: bytes) -> bool:
+        """
+        Creates the node with the specified `path` and `value`.
+
+        Auto-creates all parent nodes if they don't exist.
+
+        Does nothing if the node did not already exist.
+
+        Returns `True` if the node was created
+        """
+        try:
+            await self.create(path, value)
+            return True
+        except NodeExistsError:
+            return False
+
     async def create(self, path: str, value: bytes) -> None:
         """
         Creates the node with the specified `path` and `value`.
--- a/astacus/coordinator/plugins/zookeeper.py
+++ b/astacus/coordinator/plugins/zookeeper.py
@@ -3,13 +3,14 @@ Copyright (c) 2021 Aiven Ltd
 See LICENSE for details
 """
 from astacus.common.exceptions import TransientException
-from kazoo.client import EventType, KazooClient, KeeperState, WatchedEvent
+from kazoo.client import EventType, KazooClient, KeeperState, TransactionRequest, WatchedEvent
 from kazoo.protocol.states import ZnodeStat
 from kazoo.retry import KazooRetry
-from typing import Any, AsyncIterator, Callable, cast, Dict, List, Optional, Tuple, TypeVar
+from typing import Any, AsyncIterator, Callable, cast, Collection, Dict, List, Optional, Tuple, Type, TypeVar, Union
 
 import asyncio
 import contextlib
+import enum
 import functools
 import kazoo.exceptions
 import logging
@@ -33,6 +34,20 @@ except ImportError:
 
 
 Watcher = Callable[[WatchedEvent], None]
+
+
+class ZooKeeperTransaction:
+    def create(self, path: str, value: bytes) -> None:
+        """
+        Add a create operation to the transaction.
+        """
+        raise NotImplementedError
+
+    async def commit(self) -> None:
+        """
+        Commit the transaction.
+        """
+        raise NotImplementedError
 
 
 class ZooKeeperConnection:
@@ -82,6 +97,12 @@ class ZooKeeperConnection:
         """
         raise NotImplementedError
 
+    def transaction(self) -> ZooKeeperTransaction:
+        """
+        Begin a transaction.
+        """
+        raise NotImplementedError
+
 
 class ZooKeeperClient:
     """
@@ -106,6 +127,25 @@ class NodeExistsError(TransientException):
         self.message = f"Zookeeper node already exists: {path}"
 
 
+class TransactionError(TransientException):
+    def __init__(self, results: Collection[Union[bool, TransientException]]):
+        super().__init__(results)
+        self.message = f"Transaction failed: {results!r}"
+        self.results = results
+
+
+class RolledBackError(TransientException):
+    def __init__(self, path: str) -> None:
+        super().__init__(path)
+        self.message = f"Rolled back operation on: {path}"
+
+
+class RuntimeInconsistency(TransientException):
+    def __init__(self, path: str) -> None:
+        super().__init__(path)
+        self.message = f"Runtime inconsistency on: {path}"
+
+
 class KazooZooKeeperClient(ZooKeeperClient):
     def __init__(self, hosts: List[str], timeout: float = 10):
         self.hosts = hosts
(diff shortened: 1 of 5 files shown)


astacus/coordinator/plugins/zookeeper.py
```diff
@@ -117,6 +145,19 @@ def connect(self) -> ZooKeeperConnection:
...
 
+class KazooZooKeeperTransaction(ZooKeeperTransaction):
+    def __init__(self, request: TransactionRequest):
+        self.request = request
+
+    def create(self, path: str, value: bytes) -> None:
+        self.request.create(path, value)
+
+    async def commit(self) -> None:
+        results = await to_thread(self.request.client.retry, self.request.commit)
```
alanfranz-at-work, line 156: Is this crucial piece of code tested somewhere?

ghost (commented):
I'm +0 about this with just a couple of doubts:
* why are fake zookeeper implementations kept in the main source code, instead of the test package?
* Why are we testing fake implementations? In test_zookeeper, there're tests like ```test_fake_zookeeper_transaction_is_atomic``` that just verify fake implementations. Is it to make sure that they're working before using the fake implementations in other tests?

kmichel-aiven:
> I'm +0 about this with just a couple of doubts:
> 
>     * why are fake zookeeper implementations kept in the main source code, instead of the test package?
I believe that a "good" module should be provided with tooling to use it and develop with it, the tooling include interfaces and fakes to write unit tests against its public interface easily. (If the zookeper.py file were a separate package on pypi, and I was a user of it, I would be extremely happy to see it come with that fake interface already made, polished and reliable instead of having to implement mine because the official one is non-existent or hidden outside of the packaged module).

Plus it's not client-side javascript so I don't mind shipping a few more lines of code :)

>     * Why are we testing fake implementations? In test_zookeeper, there're tests like `test_fake_zookeeper_transaction_is_atomic` that just verify fake implementations. Is it to make sure that they're working before using the fake implementations in other tests?
Exactly that. It's safe to write unit tests against the fake only if the fake is correct, so I test the fake.

tests/unit/coordinator/plugins/test_zookeeper.py
```diff
@@ -92,3 +94,44 @@ async def test_fake_zookeeper_client_get_children_watch() -> None:
...
+        await transaction.commit()
+        assert await connection.get("/path/to/key_1") == b"content"
+        assert await connection.get("/path/to/key_2") == b"content"
+
+
+@pytest.mark.asyncio
+async def test_fake_zookeeper_transaction_is_atomic():
+    client = FakeZooKeeperClient()
+    async with client.connect() as connection:
+        await connection.create("/path/to/key_2", b"old content")
```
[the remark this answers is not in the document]
kmichel-aiven: The raw connection is just there to prepare the testing situation, the thing under test is only the "second" transaction that needs pre-existing data to be able to fail.

I vaguely prefer not using a transaction for the preparation, this way the test has less opportunities to fail for other reasons than the main topic of the test.

About serializable etc. Zookeeper transaction are just atomic bundles of operations, but they are entirely sent to the server at the "commit" step, there is no question of partial visibility for other clients.
It's exposed as many function calls because the alternative would probably be a large and weakly typed array of operations and not very elegant (in a way, on the client side, a zookeeper transaction is just a builder pattern).

astacus/coordinator/plugins/zookeeper.py
kmichel-aiven: It's missing a test, I'll add it :+1:

alanfranz-at-work:
> > * why are fake zookeeper implementations kept in the main source code, instead of the test package?
> > ```
> 
> I believe that a "good" module should be provided with tooling to use it and develop with it, the tooling include interfaces and fakes to write unit tests against its public interface easily. (If the zookeper.py file were a separate package on pypi, and I was a user of it, I would be extremely happy to see it come with that fake interface already made, polished and reliable instead of having to implement mine because the official one is non-existent or hidden outside of the packaged module).

This is probably more of an academic discussion. I think I'll want to note some of those (creating a slab page shortly) and then we may take a bit of time to discuss them, I think it would be interesting, and we would ultimately agree on some "rough guidelines".

I admit I haven't checked the code thoroughly. Is the "Fake" hierarchy actually a 100% working implementation of Zookeeper* interfaces? Then, maybe, it would be enough to rename the interface to something like "SynchronizedKVStore" (or whatever name seems suitable) and rename the fake implementation to "InProcessLocalKVStore" and the "real" implementation to "ZookeeperKVStore". Then I would be 100% fine with shipping them as alternate, simple implementations. I'm just against shipping fakes - usually partially-working, quirky, reduced implementations - in production code.

I don't know if the renaming is feasible/useful right now and this is not a blocker for approval, however.

kmichel-aiven:
> > I believe that a "good" module should be provided with tooling to use it and develop with it, the tooling include interfaces and fakes to write unit tests against its public interface easily. (If the zookeper.py file were a separate package on pypi, and I was a user of it, I would be extremely happy to see it come with that fake interface already made, polished and reliable instead of having to implement mine because the official one is non-existent or hidden outside of the packaged module).
> 
> This is probably more of an academic discussion. I think I'll want to note some of those (creating a slab page shortly) and then we may take a bit of time to discuss them, I think it would be interesting, and we would ultimately agree on some "rough guidelines".

:+1: 

> I admit I haven't checked the code thoroughly. Is the "Fake" hierarchy actually a 100% working implementation of Zookeeper* interfaces? Then, maybe, it would be enough to rename the interface to something like "SynchronizedKVStore" (or whatever name seems suitable) and rename the fake implementation to "InProcessLocalKVStore" and the "real" implementation to "ZookeeperKVStore". Then I would be 100% fine with shipping them as alternate, simple implementations. I'm just against shipping fakes - usually partially-working, quirky, reduced implementations - in production code.
> 
> I don't know if the renaming is feasible/useful right now and this is not a blocker for approval, however.

It is actually 100% working :) (within the limits of my ability to avoid introducing bugs, hence the tests).

I'm not strongly against renaming, at least the "Fake" part of the name, but we must not lose the ZooKeeper heritage in the name, the end goal is still to have code that works on a real Zookeeper, not just having a generic KVStore for astacus (since the ZooKeeper that we're manipulating is also a real thing used by ClickHouse and Flink).

astacus/coordinator/plugins/zookeeper.py
kmichel-aiven: Well, good catch, there was a bug on that part :)

alanfranz-at-work:
> I'm not strongly against renaming, at least the "Fake" part of the name, but we must not lose the ZooKeeper heritage in the name, the end goal is still to have code that works on a real Zookeeper, not just having a generic KVStore for astacus (since the ZooKeeper that we're manipulating is also a real thing used by ClickHouse and Flink).

I disagree on the naming part but, naming is one of the most difficult things to get right across a codebase and a team, so, totally expecting it :-) if it's not using zk, neither the interface nor the basic implementation should carry that name IMHO. We'll add this to the discussion points.

ghost (approved):
LGTM now.

Tests that require clickhouse are skipped, is this by design and we run them separately @kmichel-aiven ?

kmichel-aiven:
> Tests that require clickhouse are skipped, is this by design and we run them separately @kmichel-aiven ?
Installing a ClickHouse package is missing from the GitHub actions (locally you can install it and the run the tests).

PR #73, opened 2022-04-27
rohmu: Replace pghoard.rohmu with rohmu library

dmitry-potepalov:
Drop the pghoard dependency, depend only on the library itself.

### DIFF
--- a/requirements.testing.txt
+++ b/requirements.testing.txt
@@ -2,6 +2,7 @@
 pre-commit>=2.2.0
 
 # pre-commit tasks in Makefile need these
+anyio==2.0.2
 mypy==0.910
 pylint==2.9.6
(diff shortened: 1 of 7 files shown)


requirements.testing.txt
```diff
@@ -2,6 +2,7 @@
 pre-commit>=2.2.0
 
 # pre-commit tasks in Makefile need these
+anyio==3.5.0
```
fingon, line 5: F34 has 2.0.2 and F35 has 3.2; this is perhaps too new. I would suggest seeing if F34 version works as start.

fingon (changes requested):
Otherwise LGTM, but one too recent package in requirements.testing.txt.

fingon (approved):

PR #74, opened 2022-05-04
coordinator: Convert to using dependency injection and f35/aarch64 fixes

fingon:
Follow-up to last year's misc refactoring PR: https://github.com/aiven/astacus/pull/58#issuecomment-933263116

Also some aarch64/f35 fixes (use python 3.10, deprecate python 3.7, make things work with python 3.10, skip Clickhouse stuff on aarch64 as it does not currently work).

See commits for details if interested.


kmichel-aiven (dismissed):
There is an issue similar to the `AsyncSleeper` in `BackupOp` and `RestoreOp` that only appears when running ClickHouse tests.

The main cause is the fastapi choice of running all dependency injectors in a separate thread:  https://github.com/tiangolo/fastapi/blob/9090c771eea9a51b17d0eb76dd060edd9f2e640e/fastapi/dependencies/utils.py#L523-L526

See this branch-of-branch commit for a potential fix: https://github.com/aiven/astacus/commit/53db86246c05758083ce80a0c0d4b99928cb78ef (Here is a [PR](https://github.com/aiven/astacus/pull/75) in case you want to integrate it as-is)

kmichel-aiven (approved):

PR #76, opened 2022-05-10
Miscellaneous fixes

fingon:
- .pylintrc: Some tuning
- GHA tuning
- TODO: Reporting was fixed earlier
- common: Add open_path_with_atomic_rename to utils and use it in node download
- mypy.ini: Added h11 as ignored
- setup: Some requirements tuning

See commits for more details

made cassandra driver optional -> PRs should be much faster until we get cassandra really in

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -14,9 +14,6 @@ name: "CodeQL"
 on:
   push:
     branches: [ master ]
-  pull_request:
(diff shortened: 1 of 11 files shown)


fingon:
Actually I'll park this until https://github.com/aiven/astacus/pull/73 goes in - it seems that rohmu will have explicit default cloud provider dependencies, but pghoard(.rohmu) does not.

PR #77, opened 2022-05-11
Cassandra support - 2022 edition

fingon:
First draft of Cassandra plugin + some odds and ends.

This replaces https://github.com/aiven/astacus/pull/55 

Note: I changed the PR workflow to have two modes of running - minimal (base astacus) to ensure e.g. imports don't break for stuff not supported, and maximal (install whatever we can) for maximum coverage. Minimal is run only for one Python version as it is most likely not breaking differently in different Python versions, and maximal for all supported ones.

Also added codecov.io support for Astacus as I think we've chosen that as the service to use for OSS projects.

### DIFF
--- a/astacus/common/cassandra/schema.py
+++ b/astacus/common/cassandra/schema.py
@@ -1,0 +1,356 @@
+"""
+Copyright (c) 2021 Aiven Ltd
+See LICENSE for details
+
+Schema-related parts are based on basebackup_schema.py of Cashew
+
+"""
+
+from .client import CassandraSession
+from .utils import is_system_keyspace
+from astacus.common.utils import AstacusModel
+from cassandra import metadata as cm
+from typing import Any, Iterator, List, Set
+
+import hashlib
+import itertools
+import logging
+import re
+
+logger = logging.getLogger(__name__)
+
+# https://cassandra.apache.org/doc/trunk/cassandra/cql/definitions.html
+UNQUOTED_IDENTIFIER_RE = "[a-zA-Z][a-zA-Z0-9_]*"
+QUOTED_IDENTIFIER_RE = '"(?:[^"]|"")+"'
+IDENTIFIER_RE = f"(?:{UNQUOTED_IDENTIFIER_RE}|{QUOTED_IDENTIFIER_RE})"
+
+
+class CassandraNamed(AstacusModel):
+    """~abstract baseclass for something that has a name, and has cql for creating itself."""
+
+    name: str
+    cql_create_self: str
+
+    def __lt__(self, o: "CassandraNamed") -> bool:
+        # Almost all object names are unique in Cassandra, with
+        # exception of functions (see below)
+        return self.name < o.name
+
+    @classmethod
+    def from_cassandra_metadata(cls, metadata: Any) -> Any:
+        return cls(name=metadata.name, cql_create_self=metadata.as_cql_query())
+
+    def restore_if_needed(self, cas: CassandraSession, existing_map: dict) -> None:
+        if self.name in existing_map:
+            logger.debug("%r with name %r already exists", self.__class__.__name__, self.name)
+            return
+        logger.info("Creating %r %r", self.__class__.__name__, self.name)
+        cas.execute_on_all(self.cql_create_self)
+
+
+def _iterate_identifiers_in_cql_type_definition(definition: str) -> Iterator[str]:
+    # parsing might be more proper, but this is sufficient
+    for identifier in re.findall(IDENTIFIER_RE, definition):
+        if identifier.startswith('"'):
+            # quoted identifier
+            unquoted = identifier[1:-1].replace('""', '"')
+            yield unquoted
+        else:
+            # this lower is spurious, but here for
+            # understandability - Cassandra normalizes
+            # unquoted identifiers by default
+            normalized = identifier.lower()
+
+            # these are picked from https://cassandra.apache.org/doc/trunk/cassandra/cql/types.html
+            if normalized not in {"map", "set", "list", "frozen", "tuple"}:
+                yield normalized
+
+
+class CassandraUserType(CassandraNamed):
+    field_types: List[str]
+
+    @classmethod
+    def from_cassandra_metadata(cls, metadata: cm.UserType) -> "CassandraUserType":
+        return cls(name=metadata.name, cql_create_self=metadata.as_cql_query(), field_types=metadata.field_types)
+
+    def referred_normalized_identifiers(self) -> Set[str]:
+        def iterate_identifiers() -> Iterator[str]:
+            for field_type in self.field_types:
+                yield from _iterate_identifiers_in_cql_type_definition(field_type)
+
+        return set(iterate_identifiers())
+
+    def restore_in_keyspace(self, cas: CassandraSession, keyspace_metadata: cm.KeyspaceMetadata) -> None:
+        self.restore_if_needed(cas, keyspace_metadata.user_types)
+
+
+class CassandraFunction(CassandraNamed):
+    argument_types: List[str]
+
+    def __lt__(self, o: CassandraNamed) -> bool:
+        assert isinstance(o, CassandraFunction)
+        # Cassandra can have multiple functions with same name, as
+        # long as argument types differ. So use them for sorting too.
+        if self.name < o.name:
+            return True
+        if self.argument_types < o.argument_types:
+            return True
+        return False
+
+    @classmethod
+    def get_create_statement(cls, metadata: cm.Function) -> str:
+        """
+        Based on Function.as_cql_query, but doesn't strip `frozen` from arguments as they need to be specified on restoration
+        With default cassandra-driver the function fails to create with an error message like "Non-frozen UDTs are not
+        allowed inside collections: ..."
+        """
+        sep = " "
+        keyspace = cm.protect_name(metadata.keyspace)
+        name = cm.protect_name(metadata.name)
+        arg_list = ", ".join(
+            "{} {}".format(cm.protect_name(arg_name), arg_type)
+            for arg_name, arg_type in zip(metadata.argument_names, metadata.argument_types)
+        )
+        typ = metadata.return_type
+        lang = metadata.language
+        body = metadata.body
+        on_null = "CALLED" if metadata.called_on_null_input else "RETURNS NULL"
+
+        return (
+            f"CREATE FUNCTION {keyspace}.{name}({arg_list}){sep}"
+            f"{on_null} ON NULL INPUT{sep}"
+            f"RETURNS {typ}{sep}"
+            f"LANGUAGE {lang}{sep}"
+            f"AS $${body}$$"
+        )
+
+    @classmethod
+    def from_cassandra_metadata(cls, metadata: cm.Function) -> "CassandraFunction":
+        return cls(
+            name=metadata.name,
+            cql_create_self=cls.get_create_statement(metadata=metadata),
+            argument_types=metadata.argument_types,
+        )
+
+    def get_metadata_dict(self, keyspace_metadata: cm.KeyspaceMetadata) -> dict:
+        return keyspace_metadata.functions
+
+    def restore_in_keyspace(self, cas: CassandraSession, keyspace_metadata: cm.KeyspaceMetadata) -> None:
+        pretty_name = "{}.{}({})".format(
+            keyspace_metadata.name,
+            self.name,
+            ", ".join(self.argument_types),
+        )
+        for funcagg_metadata in self.get_metadata_dict(keyspace_metadata).values():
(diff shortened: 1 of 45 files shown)


astacus/node/cassandra.py
```diff
@@ -0,0 +1,156 @@
...
+
+            table_name, _ = table_name_and_id.rsplit("-", 1)
+
+            # This could be more efficient too; oh well.
+            keyspace_path = table_snapshot.parents[2]
+            table_paths = list(keyspace_path.glob(f"{table_name}-*"))
+            assert len(table_paths) >= 1, f"NO tables with prefix {table_name}- found in {keyspace_path}!"
+            if len(table_paths) > 1:
+                # Prefer the one that isn't table_name_and_id
+                table_paths = [p for p in table_paths if p.name != table_name_and_id]
```
dmitry-potepalov, line 115: wait - there were cases when `p.name == table_name_and_id`? I kinda thought that if you execute `CREATE TABLE` again on the fresh cluster you get a different ID...

astacus/node/cassandra.py
```diff
@@ -0,0 +1,156 @@
...
+            assert len(table_paths) == 1
```
dmitry-potepalov, line 116: maybe also assert that there is no data in the directory we're about to move stuff? This could lead to _interesting_ consequences if someone managed to create 2 versions of the same table (during a network split, for example) and we ended up with 2 `tableX-uuid` paths in the snapshot. On restore, first we move stuff from `tableX-1` in the snapshot to `tableX-...` in the data dirs, then try to move stuff from `tableX-2`... I'd prefer we'd just fail the restoration in that case, seems very unlikely to happen in production. _Probably_ Cassandra includes only the active version of the table in the snapshot, but I'm not sure.

dmitry-potepalov (commented):
overall LGTM, some questions in the comments

astacus/common/cassandra/schema.py
```diff
@@ -0,0 +1,336 @@
...
+    user_types: List[CassandraUserType]
+
+    @classmethod
+    def from_cassandra_metadata(cls, metadata: cm.KeyspaceMetadata) -> "CassandraKeyspace":
+        return cls(
+            # CassandraNamed
+            name=metadata.name,
+            cql_create_self=metadata.as_cql_query(),
+            # CassandraKeyspace
+            cql_create_all=metadata.export_as_string(),
```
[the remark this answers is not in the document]
fingon: It is mostly for debugging purposes - as API allows it, later on we can try to fix things if the recursive dump+restore we do actually has a bug in it. (Some text -> doesn't cost much to store, but may be useful in an emergency.)

astacus/common/cassandra/schema.py
```diff
@@ -0,0 +1,336 @@
...
+    @classmethod
+    def from_cassandra_metadata(cls, metadata: cm.UserType) -> "CassandraUserType":
+        return cls(name=metadata.name, cql_create_self=metadata.as_cql_query(), field_types=metadata.field_types)
+
+    def referred_lowercase_identifiers(self) -> Set[str]:
+        # parsing might be more proper, but this is sufficient
+        #
+        # Cassandra doesn't allow for use of keywords (e.g. frozen,
+        # tuple, ..) as identifiers, so having them returned here is
+        # fine as they won't be dependencies we really care about
```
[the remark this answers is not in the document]
fingon: It isn't as good as I thought, though.. as it turns out, as quoted identifier you can use them so this got bit more complex in the end after all.

astacus/common/cassandra/schema.py
```diff
@@ -0,0 +1,336 @@
...
+
+import hashlib
+import itertools
+import logging
+import re
+
+logger = logging.getLogger(__name__)
+
+# https://cassandra.apache.org/doc/trunk/cassandra/cql/definitions.html
+IDENTIFIER_RE = re.compile("[a-zA-Z][a-zA-Z0-9_]*")
```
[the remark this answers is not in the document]
fingon: This was actually my bad, adjusted it to work with both - the lowercasing in general was actually wrong, as it works only for unquoted identifiers (which we use in general) but with quoted identifiers you can have types `foo` and `"fOo"` and the latter is really fOo.

astacus/common/cassandra/utils.py
```diff
@@ -0,0 +1,21 @@
...
+Copyright (c) 2021 Aiven Ltd
+See LICENSE for details
+
+Miscellaneous utilities related to CassandraSession
+
+"""
+
+_SYSTEM_KEYSPACES = {
+    "system",
+    "system_auth",
```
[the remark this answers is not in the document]
fingon: History, Cashew used to do this but it seems you changed it to include it (this code is from early 2021 for most part). Guess I can remove it

astacus/common/cassandra/schema.py
```diff
@@ -0,0 +1,336 @@
...
+
+    def restore_post_data(self, cas: CassandraSession):
+        """Second part of Cassandra schema restoration.
+
+        See restore_pre_data for the flow.
+        """
+        for keyspace in self.keyspaces:
+            # Omitted for now, assume the system was configured with
+            # proper replication
+            if is_system_keyspace(keyspace):
```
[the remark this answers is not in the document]
fingon: Probably just bad comment (can't remember), but in general we don't want to restore system keyspaces so I'll just remove the comment.

The assumption is that system keyspaces get configuration/state from elsewhere than the restore process. Reworded the comment.

astacus/node/cassandra.py
```diff
@@ -0,0 +1,156 @@
...
+                ipc.CassandraSubOp.start_cassandra: self.start_cassandra,
+                ipc.CassandraSubOp.stop_cassandra: self.stop_cassandra,
+                ipc.CassandraSubOp.take_snapshot: self.take_snapshot,
+            }[subop],
+        )
+
+    def _get_schema_hash(self):
+        with CassandraClient(self.config.cassandra.client).connect() as cas:
+            schema = CassandraSchema.from_cassandra_session(cas)
+            return schema.calculate_hash()
```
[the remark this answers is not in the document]
fingon: Yes, but the hash calculation is (almost) one-liner and at least then it matches exactly what we want to restore/backup; I'm not sure which is preferable. Also, Cashew did it this way; we can easily enough change it though if you think the official schema is better though.

astacus/node/cassandra.py
fingon: This is looking at the snapshot (and table_name_and_id is from snapshot too). In some cases it seems same table name can have more than one subdirectory; not sure if this logic is copied from Cashew or new though.

astacus/node/cassandra.py
fingon: See above - I think it may include two in some cases.

fingon: I suppose that's fair - there seems to be backups/ subfolder by default but otherwise it should be empty when we start restoring it.

astacus/node/cassandra.py
fingon: I don't know what is the history behind this either anymore :-) This is actually probably artifact of restoring stuff locally to same cluster.

What is your suggestions on what to do with it - get rid of it?

astacus/common/cassandra/schema.py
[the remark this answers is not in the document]
fingon: (You can also have type called `"frozen"`)

astacus/coordinator/plugins/base.py
```diff
@@ -66,7 +66,8 @@ def get_result(self, step_class: Type[Step[T]]) -> T:
 
     def set_result(self, step_class: Type[Step[T]], result: T) -> None:
         if step_class in self.step_results:
-            raise RuntimeError(f"result already set for step {step_class}")
+            if self.step_results[step_class] is not None or result is not None:
+                raise RuntimeError(f"result already set for step {step_class}")
```
kmichel-aiven, line 70: If I understand correctly, this is linked to `CassandraSubOpStep`.
I think it would be better to be a bit more verbose and create one short subclass for each op
instead of introducing an exception to the general behavior.

This will also make the progress logging more clear, since it relies on class name.

kmichel-aiven (changes requested):
It's mostly a review about the astacus/python side of things,
I don't know Cassandra enough to comment on the backup requirements :)

Would be cool to add parameters and return type hints on all the new functions too (didn't put it on a specific line since there are a lot).

astacus/coordinator/plugins/base.py
fingon: I tried it first, but it looks pretty ugly. We're also doing CassandraStart twice, and some other things.

fingon: Having ..Again classes looks pretty lame. If progress reporting relies on this, I'd rather just use index in the steps for that primarily.

astacus/common/cassandra/utils.py
```diff
@@ -0,0 +1,21 @@
...
+    "system_distributed",
+    "system_schema",
+    "system_traces",
+}
+
+
+def is_system_keyspace(keyspace):
```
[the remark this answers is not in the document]
fingon: tbh typing this that way is painful due to circular import loop, guess I can just make it always str though.

astacus/coordinator/plugins/base.py
kmichel-aiven: It's more than reporting.

I think keeping a zero-exception "each step has its class" rule has value:
 - rules without exception are easier to explain and reason about
 - communication between people during failure diagnosis is more straightforward if the-class-is-the-step ("the second StartCassandraStep got stuck, I mean the step 8, starting at zero, except after astacus 0.7 where it's step 9" vs "RestartCassandraStep got stuck").

fingon: (or use the actual step instance and not its class name as key if we do not care about order - but I'm making order available for now)

fingon: Currently the logic where each class is essentially singleton is quite counterintuitive. The fact they're used as nice hack to get previous steps' results is mainly awkward - it would be probably preferable to have e.g. per job (series of steps) state, which is typed, and populated by the steps as they come by the state.

Insofar reporting is concerned, if log says step M of N, quoting that M is sufficient to identify step. The class names should be implementation artifact.

fingon: Possibly also refactoring if so that you can get result by type of result would make more sense as those ( if any ) seem to be unique.

astacus/common/cassandra/schema.py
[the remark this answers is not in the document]
dmitry-potepalov: it's probably affected by BF-713: if a table has a default_ttl set on it and has a materialized view, this default_ttl setting leaks into CREATE statement for the MV causing it to fail. Cashew had a hack to circumvent that, but I don't see it here. I thought that since you used a different function in metadata to capture the MV definition maybe it's not an issue anymore, but worth checking, I guess.

astacus/node/cassandra.py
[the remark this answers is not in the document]
dmitry-potepalov: since this is used to validate that schema has not changed while we're preparing the manifest, I think using the "official" version would be more appropriate. The assumption being that if the "official" version has not changed, then there were no changes we really need to care about. It also might help if half the nodes are on one version of Astacus and half on another, and some order-influencing change occured that prevent the hashes to match, while the contents of the schema remain the same.

astacus/node/cassandra.py
dmitry-potepalov: just checked, Cassandra does not clean up old table directory when you do DROP TABLE. Odd. So the multiple dir case is more common than I thought, I though you could get that only as a race condition when creating from different nodes or something.

So we need to be more clever, unfortunately. The table ID is in `system.schema`:
```
aiven@cqlsh> SELECT id FROM system_schema.tables WHERE keyspace_name='foo' AND table_name='bar';

 id
--------------------------------------
 77567990-e801-11ec-8abe-e178ff91d9b8
```
but not in client metadata, it seems. Oh well. We can include table ID in the manifest and restore data from the version active at the time we took the backup.

dmitry-potepalov (dismissed):
As discussed in comments:
* use schema_version instead of manually calculated hash
* smarter handling of the "multiple tables with the same name" case

astacus/node/cassandra.py
fingon: By default there are none though - the restore to non-empty case is simply something for later. Note: If it is not in metadata, it is quite awkward to rely on it as I think the content of system tables is not guaranteed to stay same across versions.

astacus/node/cassandra.py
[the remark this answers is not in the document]
fingon: That's fair - changed to use the official hash.

dmitry-potepalov (approved):

PR #79, opened 2022-05-25
hot-reload of configuration through API and CLI client

kmichel-aiven:
This allows updating the configuration without restarting astacus and interrupting the service.

Compared to a full restart, this removes the risk of long or leaked asyncio tasks not letting astacus restart quickly.

The reload "signal" uses an API endpoint instead of a `SIGHUP` because we can wait for its completion.
This makes it more reliable [when used with systemd](https://www.freedesktop.org/software/systemd/man/systemd.service.html#ExecReload=), or when we want to make sure a backup started
after the restart signal is really using the new configuration.

In the same spirit, an extra endpoint can tell whether the configuration needs to be reloaded.
This enables closed loop control of astacus and makes it possible to reliably run with the new configuration
even if the system in charge of reloading astacus temporarily fails or loses its state.

The configuration status endpoint shows only the hash of the loaded configuration.
This allows comparison with a know configuration file without leaking the entire configuration through the API,
which would be problematic since it contains secrets.

### DIFF
--- a/astacus/coordinator/api.py
+++ b/astacus/coordinator/api.py
@@ -35,15 +35,6 @@ class OpName(str, Enum):
     cleanup = "cleanup"
 
 
-@router.get("/{op_name}/{op_id}")
-def op_status(*, op_name: OpName, op_id: int, c: Coordinator = Depends()):
-    op, op_info = c.get_op_and_op_info(op_id=op_id, op_name=op_name)
-    result = {"state": op_info.op_status}
-    if isinstance(op, (BackupOp, RestoreOp)):
-        result["progress"] = op.progress
-    return result
-
-
 class LockStartResult(Op.StartResult):
     unlock_url: str
 
@@ -115,6 +106,15 @@ async def _list_backups(*, req: ipc.ListRequest = ipc.ListRequest(), c: Coordina
 async def cleanup(*, op: CleanupOp = Depends(), c: Coordinator = Depends()):
     runner = await op.acquire_cluster_lock()
     return c.start_op(op_name=OpName.cleanup, op=op, fun=runner)
+
+
+@router.get("/{op_name}/{op_id}")
+def op_status(*, op_name: OpName, op_id: int, c: Coordinator = Depends()):
+    op, op_info = c.get_op_and_op_info(op_id=op_id, op_name=op_name)
+    result = {"state": op_info.op_status}
+    if isinstance(op, (BackupOp, RestoreOp)):
+        result["progress"] = op.progress
+    return result
 
 
 @router.put("/{op_name}/{op_id}/sub-result")
--- a/astacus/coordinator/api.py
+++ b/astacus/coordinator/api.py
@@ -11,6 +11,7 @@ from .state import CachedListResponse
 from astacus import config
 from astacus.common import ipc
 from astacus.common.op import Op
+from astacus.config import APP_HASH_KEY, get_config_content_and_hash
 from asyncio import to_thread
 from enum import Enum
 from fastapi import APIRouter, Depends, HTTPException, Request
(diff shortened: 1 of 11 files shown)


astacus/common/compatibility.py
```diff
@@ -0,0 +1,24 @@
+"""
+Copyright (c) 2022 Aiven Ltd
+See LICENSE for details
+"""
+
+__all__ = ("to_thread",)
```
fingon, line 6: TBH this isn't really desirable to me - I have had in my Cassandra dev branch 3.9 only already for a bit, and I think we want to go there rather than add workarounds (3.9 isn't super new either). I made PR out of that now: https://github.com/aiven/astacus/pull/80

astacus/coordinator/api.py
```diff
@@ -32,25 +36,40 @@ class OpName(str, Enum):
...
+@router.post("/config/reload")
+async def config_reload(*, request: Request, c: Coordinator = Depends()):
+    """Reload astacus configuration"""
+    # First reload the config of the entire app (including coordinator and node)
+    config_path = os.environ.get("ASTACUS_CONFIG")
+    assert config_path is not None
+    config.set_global_config_from_path(request.app, config_path)
+    # Then clear the backup list cache
+    c.state.cached_list_response = None
+    if c.state.cached_list_running:
```
fingon, line 58: This seems unnecessary. Couldn't the list backups endpoint just remember the config hash it started with, and check that at the end?

fingon (dismissed):
I think that one bool flag is unnecessary, and I'd rather just upgrade to Python 3.9, but otherwise LGTM.

astacus/coordinator/api.py
kmichel-aiven: It could and that would be more elegant, I'll change it, I don't like that bool either.

kmichel-aiven: In the end, I made the list backup endpoint remember the coordinator config instead of the hash of the source file.
Not a huge observable difference (mainly, the cache is busted less often, it's not busted if the config is equivalent).

The hash is useful in the "global" part and in the externally visible API because the external user deals with the config file.
But inside the list endpoint, we only care about the actual config and using the full hash was creating an annoying dependency forcing the app fixture for coordinator tests to add a fake hash.

astacus/coordinator/api.py
```diff
@@ -75,20 +88,27 @@ async def restore(*, c: Coordinator = Depends(), op: RestoreOp = Depends(Restore
...
+        raise HTTPException(status_code=429, detail="Already caching list result")
+    c.state.cached_list_running = True
+    list_response = await to_thread(list_backups, req=req, json_mstorage=c.json_mstorage)
+    c.state.cached_list_response = CachedListResponse(
+        coordinator_config=coordinator_config,
+        list_request=req,
+        list_response=list_response,
+    )
+    c.state.cached_list_running = False
```
fingon, line 111:
```suggestion
    c.state.cached_list_running = True
    try:
        list_response = await to_thread(list_backups, req=req, json_mstorage=c.json_mstorage)
        c.state.cached_list_response = CachedListResponse(
            coordinator_config=coordinator_config,
            list_request=req,
            list_response=list_response,
          )
    finally:
        c.state.cached_list_running = False
```

fingon (dismissed):
I think there's tiny bugfix I'd like to see, but otherwise it seems fine (note: original code had the same problem, and AFAIK it never happened but when dealing with network, relying on all the sub-code not to throw an exception someday is bit optimistic perhaps).

astacus/coordinator/api.py
kmichel-aiven: Oh yeah, good catch.

fingon (approved):

PR #80, opened 2022-05-31
Change minimum Python version to 3.9

fingon:


kmichel-aiven (changes requested):
LGTM except the graphlib dependency in `setup.cfg` that should be removed too.

astacus/coordinator/plugins/clickhouse/dependencies.py
```diff
@@ -22,7 +22,7 @@ def tables_sorted_by_dependencies(tables: List[Table]) -> List[Table]:
     The `dependencies` attribute of each table must contain the list of
     `(database_name: str, table_name: str)` that depend on this table.
     """
-    sorter = graphlib.TopologicalSorter()
+    sorter = graphlib.TopologicalSorter()  # type: ignore
```
fingon, line 25: These could probably use some explicit type hint too, presumably given it is stdlib it should just work(tm) but apparently not. So punted it for later.

kmichel-aiven (approved):

PR #81, opened 2022-07-11
Fix rare ClickHouse backup restoration failure 

kmichel-aiven:
There was an opportunity for astacus to try to attach replicated
parts to table that did not exist in all replicas yet.

When that happened, the restoration would fail, complaining
that one of the tables did not exist.

All replicas are now synced after creating the tables and before
attaching parts. The synchronization is done by peeking into
the `log_ptr` stored in ZooKeeper by ClickHouse.


ghost (approved):
LGTM

PR #82, opened 2022-07-11
Fix Kazoo timeout during tests

kmichel-aiven:
The timeout was only 1s and leading to flaky tests, increased it to 10s.


ghost (approved):
LGTM

PR #83, opened 2022-07-12
Recreate ClickHouse databases with their original shard/replica/uuid

kmichel-aiven:
As a preliminary support to sharded database (this is not the full support yet),
recreate Replicated databases with their original shard and macros, according
to the definition found for each database.


tests/integration/coordinator/plugins/clickhouse/test_replication.py
```diff
@@ -0,0 +1,28 @@
...
+
+@pytest.mark.asyncio
+async def test_get_shard_and_replica(ports: Ports) -> None:
+    async with create_zookeeper(ports) as zookeeper:
+        async with create_clickhouse_cluster(zookeeper, ports, cluster_size=1) as clickhouse_cluster:
+            clickhouse = clickhouse_cluster.services[0]
+            client = get_clickhouse_client(clickhouse)
+            await client.execute(
+                b"CREATE DATABASE replicated_database "
+                b"ENGINE = Replicated('/clickhouse/databases/replicated_database', '{my_shard}', '{my_replica}')"
```
alanfranz-at-work, line 25: one curiosity: why do you use this kind of syntax? I was initially thinking there was a missing f-string. Everything works, but can be a bit confusing for the random reader.

ghost (approved):
one curiosity: why in some places you're using "{something}" as shard or replica name rather than a plain, non-curly-braced string? I was initially thinking there was a missing f-string. Everything works, but can be a bit confusing for the random reader. And in some later tests you actually escaped some curly braces since you're using an f-string.

Everything works fine, I'm just wondering about the reason (could be a bit confusing to the casual reader).

kmichel-aiven:
> one curiosity: why in some places you're using "{something}" as shard or replica name rather than a plain, non-curly-braced string? I was initially thinking there was a missing f-string. Everything works, but can be a bit confusing for the random reader. And in some later tests you actually escaped some curly braces since you're using an f-string.
> 
> Everything works fine, I'm just wondering about the reason (could be a bit confusing to the casual reader).

Oh, it's the syntax for macros in ClickHouse, which is confusingly similar to python f-strings.

They are critical for replicated databases: queries are replicated verbatim between nodes but we need each node to create a different replica, the magic is in having queries containing macros and macros having different values on each node.

PR #84, opened 2022-08-01
clickhouse: distribute parts according to shards

kmichel-aiven:
When building a backup manifest, we used to alter how files will be
restored by de-duplicating MergeTree table parts that were identical
on all servers of a cluster.

Now, we do the same thing but instead of de-duplicating across all
servers, we only do that within a single shard.

In general, each Replicated table can have a different sharding
configuration.

In practice, using the Replicated database engine, all tables inside
the same database will have the same sharding configuration.

We use this information to infer the sharding configuration of each
table, and then distribute parts of each tables according to that
configuration.

The restore operation is essentially oblivious to that manipulation.

As long as the cluster where the data is restored has the same
shape (same number of servers with the same sharding configuration
for each database), the data will be restored once per shard.

ClickHouse will replicate the data within each shard, like it already
did when the entire cluster was a single shard.

### DIFF
--- a/astacus/coordinator/plugins/clickhouse/parts.py
+++ b/astacus/coordinator/plugins/clickhouse/parts.py
@@ -7,9 +7,10 @@ Replicated family of table engines.
 
 This does not support shards, but this is the right place to add support for them.
 """
+from .escaping import escape_for_file_name, unescape_from_file_name
+from .manifest import Table
+from .replication import DatabaseReplica
 from astacus.common.ipc import SnapshotFile, SnapshotResult
-from astacus.coordinator.plugins.clickhouse.escaping import escape_for_file_name, unescape_from_file_name
-from astacus.coordinator.plugins.clickhouse.manifest import Table
 from pathlib import Path
 from typing import Dict, Iterable, List, Mapping, Optional, Sequence, Set, Tuple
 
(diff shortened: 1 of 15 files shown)


tests/unit/coordinator/plugins/clickhouse/test_parts.py
```diff
@@ -52,26 +56,55 @@ def create_part_from_part_files(part_files: List[SnapshotFile]) -> Part:
...
-            assert file.servers == {0, 1}
+        assert part.snapshot_files == original_files
+        assert part.total_size == 1120
+        assert part.servers == {0, 1}
+    assert other_files[0] == []
+    assert other_files[1] == []
+
+
+def test_group_files_into_parts_does_not_merge_parts_across_shards() -> None:
+    first_server_files = copy.deepcopy([*TABLE_1_PART_1, *TABLE_1_PART_2, *TABLE_1_PART_3])
```
dmitry-potepalov, line 84: is it worth it to write a test case when first_server_files != second_server_files? as far as I understand, it's sort-of covered in the `test_add_file_to_parts_fails_on_inconsistent_file` case, does it make sense to have a higher-level test for that as well?

dmitry-potepalov (commented):
Overall LGTM, couple of small things

tests/unit/coordinator/plugins/clickhouse/test_parts.py
kmichel-aiven: There are 2 different cases to tests :
 - a single part is present on multiple servers but some files within that part are missing on some server: this is very bad and means some of our assumption about how Clickhouse stores data: it's `test_add_file_to_parts_fails_on_inconsistent_file`
 - a entire part is present on one of the replica but not another (in the same shard): this is fine, the data snapshot happened before replication: this one is not tested but it should :+1:

astacus/coordinator/plugins/clickhouse/parts.py
```diff
@@ -51,20 +57,36 @@ def group_files_into_parts(
...
+    keyed_parts: Dict[PartKey, Dict[Path, PartFile]] = {}
     for server_index, server_files in enumerate(snapshot_files):
         for snapshot_file in server_files:
-            if not add_file_to_parts(snapshot_file, server_index, table_uuids, keyed_parts):
+            if not add_file_to_parts(snapshot_file, server_index, tables_replicas, keyed_parts):
                 other_files[server_index].append(snapshot_file)
-    return list(keyed_parts.values()), other_files
+    parts: list[Part] = []
+    for part_key, part_files in keyed_parts.items():
+        part_servers = check_part_replication(part_files.values())
```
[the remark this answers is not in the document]
kmichel-aiven: Definitely yes, I forgot to rename it :facepalm:

kmichel-aiven:
Comments addressed.

dmitry-potepalov (approved):

PR #89, opened 2022-10-08
clickhouse: increase CREATE/DROP DATABASE timeout and concurrency

kmichel-aiven:
With larger databases, if the first restoration fails, we need to drop a database,
this can take quite some time. Increase the default timeout to 300sec.

Since this can take some time, parallelize the operations, this can be very
beneficial on large clusters.

While we're changing this area, also parallelize the CREATE DATABASE
and increase the timeout, this should speed-up restoration on large clusters.

### DIFF
--- a/tests/unit/common/test_limiter.py
+++ b/tests/unit/common/test_limiter.py
@@ -3,7 +3,7 @@ Copyright (c) 2021 Aiven Ltd
 See LICENSE for details
 """
 
-from astacus.common.limiter import Limiter
+from astacus.common.limiter import gather_limited, Limiter
 from typing import Sequence
 
 import asyncio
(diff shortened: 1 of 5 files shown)


tests/unit/common/test_limiter.py
```diff
@@ -34,3 +34,31 @@ async def add_trace(start: str, sleep: float, stop: str):
...
+@pytest.mark.parametrize(
+    "limit,expected_trace",
+    [
+        (1, ["s1", "e1", "s2", "e2", "s3", "e3"]),
+        (2, ["s1", "s2", "e2", "s3", "e3", "e1"]),
+        (3, ["s1", "s2", "s3", "e2", "e3", "e1"]),
+    ],
+)
+@pytest.mark.asyncio
+async def test_gather_limited(limit: int, expected_trace: Sequence[str]) -> None:
```
alanfranz, line 48: interesting approach to testing the Limiter. it took a bit to understand, but I like it: no stubbing, no mocking, no exposure of internal state, only behaviour is tested.

alanfranz (approved):
LGTM

PR #90, opened 2022-10-08
cassandra: add asyncio marker for async tests

kmichel-aiven:
Tests were skipped because of the missing marker.


alanfranz:
It seems that most freshly-marked tests are still skipped though: https://github.com/aiven/astacus/actions/runs/3210107267/jobs/5247323696

only the ones in test_utils now run. Is this intended behaviour?

kmichel-aiven:
> It seems that most freshly-marked tests are still skipped though: https://github.com/aiven/astacus/actions/runs/3210107267/jobs/5247323696
> 
> only the ones in test_utils now run. Is this intended behaviour?

These are skipped for a different reason, the cassandra library is not installed, I don't know if that's intentional.

alanfranz (approved):
LGTM.

PR #91, opened 2022-11-02
Python 3.11 + Fedora 37

fingon:


fingon:
Bit premature perhaps, I'll try to get working it in local container first.

PR #92, opened 2022-11-06
clickhouse: add support and tests for ClickHouse 22.8

kmichel-aiven:
The only real new feature is supporting a different way of authenticating Replicated
databases, using a named collection (available since ClickHouse 22.4).

The rest is extending the test setup to handle minor differences in tables metadata
and giving some tools to test the transition between two different ClickHouse versions.


aris-aiven (approved):
LGTM, merging it.

PR #93, opened 2022-11-15
Update linters and typing

aris-aiven:
This PR is a maintenance update that aims at improving CI and updating requirements.

It streamlines build and test requirements, updates pre-commit hook versions, fixes the errors surfaced by pylint and mypy, and normalizes formatting for markdown and yaml according to a prettier pre-commit hook.

This is a follow up of #95 and prepares the field for an update to python 3.11.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -4,15 +4,15 @@ on: pull_request
 
 jobs:
   lint:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       - name: Install dependencies
         run: make build-dep-ubuntu
 
       - name: Set up Python 3.10 with caching
-        uses: actions/setup-python@v3
+        uses: actions/setup-python@v4
         with:
           python-version: '3.10'
           cache: 'pip'
@@ -24,9 +24,9 @@ jobs:
         run: make lint
 
   copyright:
-      runs-on: ubuntu-latest
+      runs-on: ubuntu-22.04
       steps:
-        - uses: actions/checkout@master
+        - uses: actions/checkout@v3
 
         - name: Enforce copyright headers
           run: make copyright
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -41,13 +41,15 @@ repos:
         args:
           - --rcfile=.pylintrc
 
-  # https://pre-commit.com/#repository-local-hooks
-  - repo: local
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.910
     hooks:
       - id: mypy
-        name: mypy
-        entry: mypy
-        language: system
-        types: [python]
-        files: \.py$
         exclude: ^setup.py|^tests/|^vendor/|^astacus/proto/
+        additional_dependencies:
+          - types-PyYAML>=6.0.12.2
+          - types-requests>=2.28.11.5
+          - types-tabulate>=0.9.0.0
+          - types-urllib3>=1.26.25.4
+          - typing_extensions>=3.10.0
+          - pydantic==1.7.4
--- a/astacus/common/cassandra/schema.py
+++ b/astacus/common/cassandra/schema.py
@@ -84,6 +84,12 @@ class CassandraUserType(CassandraNamed):
         self.restore_if_needed(cas, keyspace_metadata.user_types)
 
 
+def _get_field_repr(arg_name_type: tuple[str, str]) -> str:
+    arg_name, arg_type = arg_name_type
+    protected_name = cm.protect_name(arg_name)
+    return f"{protected_name} {arg_type}"
+
+
 class CassandraFunction(CassandraNamed):
     argument_types: List[str]
 
@@ -107,10 +113,8 @@ class CassandraFunction(CassandraNamed):
         sep = " "
         keyspace = cm.protect_name(metadata.keyspace)
         name = cm.protect_name(metadata.name)
-        arg_list = ", ".join(
-            "{} {}".format(cm.protect_name(arg_name), arg_type)
-            for arg_name, arg_type in zip(metadata.argument_names, metadata.argument_types)
-        )
+
+        arg_list = ", ".join(map(_get_field_repr, zip(metadata.argument_names, metadata.argument_types)))
         typ = metadata.return_type
         lang = metadata.language
         body = metadata.body
@@ -136,11 +140,8 @@ class CassandraFunction(CassandraNamed):
         return keyspace_metadata.functions
 
     def restore_in_keyspace(self, cas: CassandraSession, keyspace_metadata: cm.KeyspaceMetadata) -> None:
-        pretty_name = "{}.{}({})".format(
-            keyspace_metadata.name,
-            self.name,
-            ", ".join(self.argument_types),
-        )
+        argument_types_str = ", ".join(self.argument_types)
+        pretty_name = f"{keyspace_metadata.name}.{self.name}({argument_types_str})"
         for funcagg_metadata in self.get_metadata_dict(keyspace_metadata).values():
             if funcagg_metadata.name == self.name and funcagg_metadata.argument_types == self.argument_types:
                 logger.info("%s %s already exists, not creating it", self.__class__.__name__, pretty_name)
@@ -163,7 +164,8 @@ class CassandraAggregate(CassandraFunction):
         ret = f"CREATE AGGREGATE {keyspace}.{name}({type_list}){sep}" f"SFUNC {state_func}{sep}" f"STYPE {state_type}"
 
         if metadata.final_func:
-            ret += "{}FINALFUNC {}".format(sep, cm.protect_name(metadata.final_func))
+            final_func = cm.protect_name(metadata.final_func)
+            ret += f"{sep}FINALFUNC {final_func}"
         if metadata.initial_condition:
             ret += f"{sep}INITCOND {metadata.initial_condition}"
 
--- a/astacus/common/utils.py
+++ b/astacus/common/utils.py
@@ -8,6 +8,8 @@ See LICENSE for details
 Shared utilities (between coordinator and node)
 
 """
+from __future__ import annotations
+
 from abc import ABC
 from contextlib import contextmanager
 from multiprocessing.dummy import Pool  # fastapi + fork = bad idea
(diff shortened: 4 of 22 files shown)


aris-aiven:
This will be rebased off #95 and the python 3.11 commits moved to another PR.

astacus/coordinator/plugins/clickhouse/steps.py
```diff
@@ -420,15 +420,17 @@ async def run_step(self, cluster: Cluster, context: StepsContext) -> None:
                 for client in self.clients
             ],
         )
+
+        def _escape_if_string(value):
```
kmichel-aiven, line 424: Missing parameter and return type hint.

tests/unit/coordinator/plugins/clickhouse/test_parts.py
```diff
@@ -119,7 +118,7 @@ def test_group_files_into_parts_ignores_parts_from_unselected_tables() -> None:
     second_server_files = copy.deepcopy([*TABLE_3_PART_1B])
     tables_replicas = {T1_UUID: DATABASES_REPLICAS, T2_UUID: DATABASES_REPLICAS}
     parts, other_files = group_files_into_parts([first_server_files, second_server_files], tables_replicas)
-    assert parts == []
+    assert not parts
```
kmichel-aiven, line 121: This, and all the other similar fixes, I assume it's because of the `use-implicit-booleaness-not-comparison` pylint warning, but I think this is a wrong hint and can lead to code that isn't clear about its intent and even bugs:

- you have a test that check that `function_x` returns an empty list
- the linters tell you to booleanize the test
- the `function_x` is modified to return an `list[T] | None` instead of `list[T]`
- now your test is confusing empty list and no list at all, if `function_x` is returning something like "an optional list of allowed things", then the difference is very important.


I'd prefer to disable this lint in .pylintrc.

tests/unit/common/test_utils.py
```diff
@@ -155,18 +155,18 @@ def test_open_path_with_atomic_rename(tmpdir):
     f1_path = f"{tmpdir}/f1"
     with utils.open_path_with_atomic_rename(f1_path) as f1:
         f1.write(b"test1")
-    assert Path(f1_path).read_text() == "test1"
+    assert Path(f1_path).read_text(encoding="utf-8") == "test1"
```
kmichel-aiven, line 158: I think it's because pylintrc `unspecified-encoding` warning, but this has only value for cross platform code.
I don't think we care about that for astacus, we could just disable the warning entirely.

kmichel-aiven (changes requested):

astacus/coordinator/plugins/clickhouse/steps.py
aris-aiven: Extracting it and placing it below the imports

tests/unit/coordinator/plugins/clickhouse/test_parts.py
aris-aiven: Agree the implicit booleaness leads to ambiguous code. Disabling the warning and reverting these.

.pre-commit-config.yaml
```diff
@@ -41,13 +41,20 @@ repos:
         args:
           - --rcfile=.pylintrc
 
-  # https://pre-commit.com/#repository-local-hooks
-  - repo: local
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.991
```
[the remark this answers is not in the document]
aris-aiven: I'll pin both requirements to the same version.
If one only runs it using `pre-commit` then the mypy version in the requirements is not needed, but I expect it's useful to have mypy available in the dev venv.

tests/unit/common/test_utils.py
aris-aiven: Done

fingon (changes requested):
Please don't apply arbitrary reformatting. If you do, and we agree to do it, let's agree on formatting change and do that ONLY in a single PR. This PR both changes formatting AND changes content, and that is painful to review.

Disclaimer: partial review follows.

.pre-commit-config.yaml
```diff
@@ -47,11 +47,11 @@ repos:
       - id: mypy
         exclude: ^setup.py|^tests/|^vendor/|^astacus/proto/
         additional_dependencies:
-          - types-PyYAML>=6.0.12.2
-          - types-requests>=2.28.11.4
-          - types-tabulate>=0.9.0.0
-          - types-urllib3>=1.26.25.3
-          - typing_extensions>=3.10.0
+          - types-PyYAML
```
fingon, line 50: Why drop explicit versions if they are required?

.pre-commit-config.yaml
```diff
@@ -41,13 +41,20 @@ repos:
...
-        types: [python]
-        files: \.py$
         exclude: ^setup.py|^tests/|^vendor/|^astacus/proto/
+        additional_dependencies:
+          - types-PyYAML>=6.0.12.2
+          - types-requests>=2.28.11.4
+          - types-tabulate>=0.9.0.0
+          - types-urllib3>=1.26.25.3
+          - typing_extensions>=3.10.0
+          - pydantic<2.0.0,>=1.6.2
```
fingon, line 55: The version used here should be exact match with what we use elsewhere.

aris-aiven:
Well noted regarding the formatting, will draft another PR so it can be discussed first (maybe it's too low priority).

.pre-commit-config.yaml
aris-aiven: Thanks for catching that

.pre-commit-config.yaml
aris-aiven: Will double check that the constraints are correct.

aris-aiven:
Formatting moved to separate PR #98

.pre-commit-config.yaml
aris-aiven: Doubled checked and everything's pinned to the same versions.

.github/workflows/lint.yml
```diff
@@ -4,15 +4,15 @@ on: pull_request
 
 jobs:
   lint:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
fingon, line 7: This is probably not desirable - the Dockerfile should match the version, and I don't think github and docker repos' definition of latest is same. (Not repeating it over later points.)

.pre-commit-config.yaml
```diff
@@ -41,13 +41,15 @@ repos:
...
-        types: [python]
-        files: \.py$
         exclude: ^setup.py|^tests/|^vendor/|^astacus/proto/
+        additional_dependencies:
+          - types-PyYAML>=6.0.12.2
+          - types-requests>=2.28.11.5
+          - types-tabulate>=0.9.0.0
+          - types-urllib3>=1.26.25.4
+          - typing_extensions>=3.10.0
+          - pydantic<2.0.0,>=1.7.4
```
fingon, line 55: This should be 1.7.4 (f35) and whatever (once we get to f37).

astacus/common/cassandra/schema.py
```diff
@@ -108,7 +108,7 @@ def get_create_statement(cls, metadata: cm.Function) -> str:
         keyspace = cm.protect_name(metadata.keyspace)
         name = cm.protect_name(metadata.name)
         arg_list = ", ".join(
-            "{} {}".format(cm.protect_name(arg_name), arg_type)
+            f"{cm.protect_name(arg_name)} {arg_type}"
```
fingon, line 111: (megacorp policy) no functions in f-strings please.

astacus/common/utils.py
```diff
@@ -274,11 +276,11 @@ def timedelta_as_short_str(delta):
...
 
-def size_as_short_str(s):
-    for u, m in [("T", 1e12), ("G", 1e9), ("M", 1e6), ("K", 1e3)]:
-        if s >= m:
-            return "%.1f %sB" % (s / m, u)
-    return f"{s} B"
+def size_as_short_str(size: int | float) -> str:
+    for unit, multiplier in [("T", 1e12), ("G", 1e9), ("M", 1e6), ("K", 1e3)]:
+        if size >= multiplier:
+            return f"{size / multiplier:.1f} {unit}B"
```
fingon, line 282: (megacorp policy) No math in f-strings please either.

astacus/coordinator/plugins/clickhouse/steps.py
```diff
@@ -420,15 +427,14 @@ async def run_step(self, cluster: Cluster, context: StepsContext) -> None:
                 for client in self.clients
             ],
         )
+
         settings = [
-            "{}={}".format(
-                setting_name,
-                escape_sql_string(value.encode()) if isinstance(value, str) else value,
-            )
+            f"{setting_name}={_escape_if_string(value)}"
```
fingon, line 432: f-string.. ;)

fingon (changes requested):
Much better with formatting elsewhere. Some comments (primarily on f-string megacorp rules, sigh).

.github/workflows/lint.yml
aris-aiven: Right, didn't think about the Dockerfile. Bumping both to 22.04!

.pre-commit-config.yaml
aris-aiven: Fixing the constraint

astacus/common/utils.py
aris-aiven: Agree it's slower and less clear to read.

astacus/common/cassandra/schema.py
aris-aiven: Untangling that

astacus/coordinator/plugins/clickhouse/steps.py
aris-aiven: Ack

.github/workflows/lint.yml
aris-aiven: Oddly enough, `latest` means `20.04` still (https://github.com/actions/runner-images)

.pre-commit-config.yaml
aris-aiven: Pinning pydantic to `1.7.4` clashes with `plugins = pydantic.mypy` in `mypy.ini` due to a breaking change in mypy `0.920` (see https://github.com/pydantic/pydantic/issues/3528). The first pydantic version that fixes this is `1.9.1`.
IMO pinning mypy to `0.910` seems the most reasonable thing to do here.

astacus/common/cassandra/schema.py
```diff
@@ -97,6 +97,12 @@ def __lt__(self, o: CassandraNamed) -> bool:
             return True
         return False
 
+    @staticmethod
```
fingon, line 100: nit: staticmethods should be used sparingly (as this doesn't really need the class or instance, it could be just separate function)

mypy.ini
```diff
@@ -1,5 +1,5 @@
 [mypy]
-python_version = 3.9
+python_version = 3.10
```
fingon, line 2: Typically we've used mypy with oldest version, so that if someone uses tooling which depends on types in the oldest supported version it works out. Due to that, as we still plan to support 3.9, this doesn't feel like good idea (might invalidate some of the changes too).

fingon (changes requested):
Mostly LGTM, but I think this contains bits and pieces of the other PR too (the new statistics value). Perhaps rebase will fix it?

astacus/common/cassandra/schema.py
aris-aiven: Done

mypy.ini
aris-aiven: Right. Changed it while debugging some mypy failures, reverted now.

aris-aiven:
The backup size estimation had this PR set as base branch, so when it was merged first GitHub got confused I guess. Rebasing.

fingon (approved):

PR #94, opened 2022-11-15
Fix clickhouse version CLI output parsing

aris-aiven:
Conftest ClickHouse version command output parsing fails for the latest official LTS build.

### DIFF
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -1,0 +1,19 @@
+"""
+Copyright (c) 2022 Aiven Ltd
+See LICENSE for details
+"""
+from collections.abc import Sequence
+from tests.utils import parse_clickhouse_version
+
+import pytest
+
+
+@pytest.mark.parametrize(
+    "version_command_output,expected_version",
+    [
+        (b"ClickHouse server version 22.0.0.1 (official build).", (22, 0, 0, 1)),
+        (b"ClickHouse server version 22.1.2.3.", (22, 1, 2, 3)),
+    ],
+)
+def test_parse_clickhouse_version(version_command_output: bytes, expected_version: Sequence[int]) -> None:
+    assert parse_clickhouse_version(version_command_output) == expected_version
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -2,9 +2,12 @@
 Copyright (c) 2020 Aiven Ltd
 See LICENSE for details
 """
-
 from astacus.common.rohmustorage import RohmuConfig
 from pathlib import Path
+from typing import List, Sequence, Union
+
+import re
+import subprocess
 
 # These test keys are from copied from pghoard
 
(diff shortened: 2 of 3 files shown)


aris-aiven:
CI is currently broken, it needs to be re-run when #93 is merged.

kmichel-aiven (dismissed):
Following an out-of-band discussion, this should include a few unit tests to show the version strings we expect to successfully parse.

tests/utils.py
```diff
@@ -68,3 +71,12 @@ def create_rohmu_config(tmpdir, *, compression=True, encryption=True):
             }
         )
     return RohmuConfig.parse_obj(config)
+
+
+def get_clickhouse_version(command: List[Union[str, Path]]) -> tuple[int]:
```
kmichel-aiven, line 76: `tuple[int]` is not a tuple of ints, it's a tuple with a single int, you need `tuple[int, ...]` or `Sequence[int]` (I think the second one is more clear, it requires less knowledge to understand what it means).

tests/unit/test_utils.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+@pytest.mark.parametrize(
+    "command,output,expected",
+    [
+        ([Path("/usr/bin/clickhouse"), "server"], b"ClickHouse server version 22.0.0.1 (official build).", (22, 0, 0, 1)),
+        ([Path("/usr/local/bin/clickhouse", "server")], b"ClickHouse server version 22.1.2.3.", (22, 1, 2, 3)),
+    ],
+)
+def test_get_clickhouse_version(command: List[Union[str, Path]], output: bytes, expected: tuple[int]):
```
kmichel-aiven, line 20: Here too, the type of of `expected` is not `tuple[int]` (and the return type is missing).

tests/unit/test_utils.py
```diff
@@ -0,0 +1,22 @@
...
+    with patch("subprocess.check_output", return_value=output):
```
kmichel-aiven, line 21: Instead of having to `mock.patch`, I think it would be better to have a `parse_clickhouse_version(command_output: bytes) -> Sequence[int]` and keep the `subprocess` outside of the parsing function.

kmichel-aiven (changes requested):

tests/utils.py
aris-aiven: You're right, somehow I thought it type-checked any homogenous tuple type.

tests/unit/test_utils.py
aris-aiven: Ack

tests/unit/test_utils.py
aris-aiven: Will do, hesitated as it seemed overkill to split that.

kmichel-aiven: I think it's fine if the subprocess goes back into the original place without creating a third function, it's not like we use it more than once.

PR #97, opened 2022-11-22
Fix backup size computation in the presence of duplicate snapshot files

aris-aiven:
When multiple nodes snapshot files with the same digest, the list operation sums their size incorrectly when returning the corresponding backup's size. 

Deduplicate the snapshot files by hexdigest, and retain the previous behaviour if the backup manifest has no associated snapshot state.

### DIFF
--- a/astacus/common/ipc.py
+++ b/astacus/common/ipc.py
@@ -125,7 +125,10 @@ class SnapshotResult(NodeResult):
     start: datetime = Field(default_factory=now)
     end: Optional[datetime]
 
-    # should be passed opaquely to restore
+    # The state is optional because it's written by the Snapshotter post-initialization.
+    # If the backup failed, the related manifest doesn't exist: the state and the
+    # summary attributes below will be set to none and their default values respectively.
+    # Should be passed opaquely to restore.
     state: Optional[SnapshotState] = Field(default_factory=SnapshotState)
 
     # Summary data for manifest use
@@ -259,6 +262,11 @@ class ListSingleBackup(AstacusModel):
     total_size: int
     upload_size: int
     upload_stored_size: int
+    # Number of cluster files and total cluster data size
+    # The two fields are computed from deduplicated files across nodes but not within nodes,
+    # using the files' hexdigest as key. As such, they are *not* sourced from BackupManifest.
+    cluster_files: int
+    cluster_data_size: int
 
 
 class ListForStorage(AstacusModel):
--- a/astacus/coordinator/list.py
+++ b/astacus/coordinator/list.py
@@ -3,12 +3,38 @@ Copyright (c) 2020 Aiven Ltd
 See LICENSE for details
 
 """
-
 from astacus.common import ipc, magic
-from astacus.common.storage import MultiStorage
+from astacus.common.storage import JsonStorage, MultiStorage
+from collections import defaultdict
+from collections.abc import Iterator
 
 
-def _iter_backups(storage):
+def compute_deduplicated_snapshot_file_stats(manifest: ipc.BackupManifest) -> tuple[int, int]:
+    """Compute stats over snapshot files as identified by their hex digest.
+
+    There may be duplicate hex digests within nodes for multiple copies of the same data chunks.
+    Duplicates' size are aggregated as they require physical disk space during restore.
+    On the other hand, hex digests can be safely deduplicated across nodes. The ocurrence with
+    the highest associated file count is kept, giving an upper bound for the cluster data size.
+    """
+    hexdigest_max_counts: dict[str, int] = {}
+    hexdigest_sizes: dict[str, int] = {}
+    for snapshot_result in manifest.snapshot_results:
+        assert snapshot_result.state is not None
+        node_hexdigest_counter: defaultdict[str, int] = defaultdict(lambda: 0)
+        for snapshot_file in snapshot_result.state.files:
+            node_hexdigest_counter[snapshot_file.hexdigest] += 1
+            if snapshot_file.hexdigest not in hexdigest_sizes:
+                hexdigest_sizes[snapshot_file.hexdigest] = snapshot_file.file_size
+        for hexdigest, count in node_hexdigest_counter.items():
+            max_count = hexdigest_max_counts.get(hexdigest, 0)
+            hexdigest_max_counts[hexdigest] = max(max_count, count)
+    num_files = sum(hexdigest_max_counts.values())
+    total_size = sum(count * hexdigest_sizes[hexdigest] for hexdigest, count in hexdigest_max_counts.items())
+    return num_files, total_size
+
+
+def _iter_backups(storage: JsonStorage) -> Iterator[ipc.ListSingleBackup]:
     for name in sorted(storage.list_jsons()):
         if not name.startswith(magic.JSON_BACKUP_PREFIX):
             continue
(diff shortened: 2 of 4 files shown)


astacus/coordinator/list.py
```diff
@@ -3,19 +3,37 @@
...
+        if not snapshot_result.state:
+            num_files += snapshot_result.files
+            total_size += snapshot_result.total_size
+        else:
+            for snapshot_file in snapshot_result.state.files:
+                if snapshot_file.hexdigest in seen_hexdigests:
+                    continue
+                seen_hexdigests.add(snapshot_file.hexdigest)
+                num_files += 1
+                total_size += snapshot_file.file_size
```
kmichel-aiven, line 26:
```suggestion
                num_files += 1
                total_size += snapshot_file.file_size
            for snapshot_file in snapshot_result.state.files:
                seen_hexdigests.add(snapshot_file.hexdigest)
```

I think we shouldn't de-duplicate within a single `snapshot_result`, while still de-duplicating between different `snapshot_result`:

We don't want to count twice the data that is identical between two different nodes because they are in the same shard (or replicas, depending on the service type of the backup).

But we also don't want the original uncompressed total size to be smaller than the volume on data on a single node, and if someone has created 10 identical 1GB tables, which are taking 10GB on the disk, we want the original size to reflect that (this will matter when estimating whether we can restore the backup on a specific disk of fixed size).

(that means the function name should also change).

tests/unit/coordinator/test_list.py
```diff
@@ -71,3 +93,207 @@ def _run():
...
+        ],
+        upload_results=[
+            SnapshotUploadResult(total_size=4000, total_stored_size=3000),
+            SnapshotUploadResult(total_size=4000, total_stored_size=3000),
+        ],
+        plugin=Plugin.cassandra,
+    )
+
+
+def test_compute_unique_snapshot_file_stats(backup_manifest: BackupManifest):
```
kmichel-aiven, line 252: nit. return type is missing (next function too).

tests/unit/coordinator/test_list.py
```diff
@@ -4,10 +4,32 @@
...
+from os import PathLike
+from pathlib import Path
+from tests.utils import create_rohmu_config
 
+import datetime
+import pytest
 
-def test_api_list(client, populated_mstorage, mocker):
+
+def test_api_list(client: TestClient, populated_mstorage: MultiRohmuStorage, mocker):
```
kmichel-aiven, line 32: nit. return type is missing.

kmichel-aiven (dismissed):

tests/unit/coordinator/test_list.py
aris-aiven: Thanks

tests/unit/coordinator/test_list.py
kmichel-aiven: I wish I could force mypy to do it, it's really not intuitive that "nothing" means "anything" instead of None.

(Which, in case of tests, becomes fun when you accidentally `return expected` instead of `assert expected` and mypy doesn't complain).

fingon (changes requested):
If you want to add separate return value, that's fine, but changing existing ones is not as e.g. prune has some logic which attempts also to estimate this. Due to that, this change is invalid, as changing existing values' behavior is not good.

tests/unit/coordinator/test_list.py
fingon: There is an option to force typing, it is not just set for astacus as I didn't have time for it, see mypy.ini:

```
#disallow_incomplete_defs = true
#disallow_untyped_defs = true
```

kmichel-aiven:
> If you want to add separate return value, that's fine, but changing existing ones is not as e.g. prune has some logic which attempts also to estimate this. Due to that, this change is invalid, as changing existing values' behavior is not good.

That's fair.
Giving it a clear name is not easy, does something like that make sense ?
 - total_size : sum of the size of all files
 - **cluster_data_size : sum of the size of all files, deduplicated between nodes but not within each node**
 - upload_size : sum of the size of all unique files, that were not part of the previous backup
 - upload_stored_size : sum of the size of all unique compressed files that were not part of the previous backup

astacus/coordinator/list.py
aris-aiven: Thank you for the comment. The function was reworked to aggregate size and file count information globally on a per-hexdigest basis and apply a reduction to the stats gathered.

tests/unit/coordinator/test_list.py
aris-aiven: IIUC enabling these will most probably main a deluge of errors in the future, whenever an untyped module is touched.

aris-aiven:
Going for the `cluster_data_size` and mirror `cluster_files` naming, let me know if you think of something more appropriate.

astacus/common/ipc.py
```diff
@@ -256,7 +256,9 @@ class ListSingleBackup(AstacusModel):
     attempt: int
     nodes: int
     files: int
+    cluster_files: int
     total_size: int
+    cluster_data_size: int
     upload_size: int
     upload_stored_size: int
```
fingon, line 263: It would be good to document here (with comments) what Kevin noted in the PR, e.g. what each of these means. Also fields from nodes onward are actually NOT from BackupManifest so they should be really documented here.

astacus/coordinator/list.py
```diff
@@ -3,12 +3,48 @@
...
+    the highest associated files' count and maximum file size across all nodes is kept.
+    """
+    global_file_counter = defaultdict(list)
+    global_file_sizes = defaultdict(list)
+    num_files = 0
+    total_size = 0
+    for snapshot_result in manifest.snapshot_results:
+        if not snapshot_result.state:
+            num_files += snapshot_result.files
+            total_size += snapshot_result.total_size
```
fingon, line 27: This is probably not necessary. `state` is optional in typing so that empty result can be initialized, but if it is not set, these values are also not set (see node/snapshot.py) and the backup has failed so manifest is unlikely to exist. Could just skip these (+- asserting state is set) and calculate `num_files` and `total_size` in the end without the initial `= 0` and the `+=`s here.

(The intermediate result without it set is only used for progress reporting and while decoupling these two would be possible, I haven't bothered to.)

This would be perhaps worth documenting in ipc.py too.

astacus/coordinator/list.py
```diff
@@ -3,12 +3,48 @@
...
+            node_hexdigest_counter: defaultdict[str, int] = defaultdict(lambda: 0)
+            node_hexdigest_sizes: defaultdict[str, list[int]] = defaultdict(list)
+            for snapshot_file in snapshot_result.state.files:
+                node_hexdigest_counter[snapshot_file.hexdigest] += 1
+                node_hexdigest_sizes[snapshot_file.hexdigest].append(snapshot_file.file_size)
+            for hexdigest, count in node_hexdigest_counter.items():
+                node_hexdigest_max_size = max(node_hexdigest_sizes[hexdigest])
+                global_file_sizes[hexdigest].append(node_hexdigest_max_size)
+                global_file_counter[hexdigest].append(count)
+    # This is not strictly correct as it takes the global max size and not the size associated with
```
fingon, line 38: size is constant (hexdigest is hash of the file -> likelihood of having same hash map to files with different sizes is very, very slim). I'm not sure if it affects anything here as such but the comment reads wrong.

tests/unit/coordinator/test_list.py
fingon: Well, those should really be enabled in some future PR which does it for whole codebase. (I had partial effort but it turned out to be bit much).  Probably starting with incomplete defs would help, and then later on doing all untyped ones.

fingon (changes requested):

astacus/common/ipc.py
aris-aiven: Points adressed

astacus/coordinator/list.py
aris-aiven: I'm ashamed haha

astacus/coordinator/list.py
aris-aiven: Removed the dead code and added a comment in `ipc`

fingon (approved):

PR #98, opened 2022-11-24
Formatting proposal relying on prettier pre-commit hook

aris-aiven:
The idea is to standardize formatting for toml, yaml, json & markdown files.

It's not strictly necessary but it's nice to have sane defaults that allow everyone to focus on the actual code.

The [prettier.io](prettier.io) solution with the accompanying pre-commit hook is widely used, although certainly not the only one. 

Please let me know if this is something that would be valuable for the repository and whether you have better or alternative solutions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ multiple open-source cluster databases, such as
 [Apache Cassandra](https://cassandra.apache.org).
 
 _My name is Maximus Backupus Astacus, Co-ordinator of the Backups of the
-Cluster, Master  of the Storage Availability, loyal servant to the true
+Cluster, Master of the Storage Availability, loyal servant to the true
 emperor, Prunus Aivenius. Father to a failed backup, husband to a corrupted
 data. And I will have my restore, in this runtime or the next._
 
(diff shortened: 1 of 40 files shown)


README.md
```diff
@@ -17,17 +17,16 @@ data. And I will have my restore, in this runtime or the next._
 - Product-specific code with simple, testable API
 
 - Complexities to deal with e.g. reuse of blobs with same value in the
-shared code
-    - It is needed to accomplish e.g. fast non-incremental M3 backups that are
```
fingon, line 21: This looks like mutilated second level list

fingon (changes requested):
One error and I am personally not that fond of the default formatting for json for example either. Question still is, why? It has nontrivial error rate probably, and value of autoformatting non-code seems somewhat slim.

If we want to go in this direction, I'd rather get company-wide consensus on formatting anyway rather than do it in this one project only (and I'm not sure how much of a buy-in this would get elsewhere).

README.md
aris-aiven: Thanks, fixed that.

PR #100, opened 2022-11-28
Reduce the amount of checksuming for ClickHouse

kmichel-aiven:
ClickHouse databases are quite large, hashing each file 3 times for each upload
consumes a significant amount of CPU time (and probably does not help the
IO cache).

Since ClickHouse files never change, only hash them during the snapshot and
not again before and after each upload.

Also increase the threshold for embedded files from 100 to 200 bytes, it's still
small but it helps to include many small ClickHouse files that are exactly
185 bytes (the checksums.txt file present for each table part).

We're adding a new flag to the `/node/upload` API, this would completely fail
on existing nodes, to deal with that, add a `/node/metadata` API that list the
node version and supported features.

For backward compatibility, assume a node not replying to that endpoint has
no feature.

### DIFF
--- a/astacus/common/ipc.py
+++ b/astacus/common/ipc.py
@@ -120,6 +120,15 @@ class SnapshotUploadRequest(NodeRequest):
     storage: str
 
 
+# Added on 2022-11-29, the previous version should be removable after 1 or 2 years.
+class SnapshotUploadRequestV20221129(SnapshotUploadRequest):
+    # Whether we should check if the file hash has changed since the snapshot
+    # Should be False only for plugins where the database is known to never
+    # change the files that we are reading: for instance because the database
+    # itself created a snapshot and Astacus is reading from that snapshot.
+    validate_file_hashes: bool = True
+
+
 class SnapshotUploadResult(NodeResult):
     total_size: int = 0
     total_stored_size: int = 0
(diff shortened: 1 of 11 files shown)


astacus/common/ipc.py
```diff
@@ -115,6 +120,14 @@ class SnapshotUploadRequest(NodeRequest):
     storage: str
 
 
+class SnapshotUploadRequestV2(SnapshotUploadRequest):
```
fingon, line 123: I would perhaps add some sort of version indicator here - e.g. date when added, or version after which (we're bad at versions, so perhaps date is better). Once long enough time has passed, we can drop the old one and associated code. (I'd say year or two.)

astacus/server.py
```diff
@@ -43,8 +43,8 @@ def init_app():
     config_path = os.environ.get("ASTACUS_CONFIG")
     assert config_path
     api = FastAPI()
-    api.include_router(coordinator_router, tags=["coordinator"])
     api.include_router(node_router, prefix="/node", tags=["node"])
+    api.include_router(coordinator_router, tags=["coordinator"])
```
fingon, line 47: nit: ordering changed to non-alphabetical, why? ;)

fingon (commented):
Overall LGTM. Nit or two, and I wonder about the general feature list deprecation but I doubt it will become a problem if we even never do that (and in some cases, like this, it can be perhaps done for code cleanup reasons after awhile).

fingon (approved):
Just V2 would've been probably enough for naming purposes but I guess this works too. (I wanted it just documented when we can get rid of it / assume it to be present).

PR #101, opened 2023-01-08
clickhouse: allow restoring tables requiring global settings

kmichel-aiven:
A table can be created using a transient global settings:
a global settings that was enabled for a single session or a single
query.

For the settings related to experimental features, if we don't
enable them again when recreating the table, the restoration
will fail.

Try to enable those settings before creating all tables:
 - If it's necessary, the table will be restorable.
 - If it's unnecessary, it will be harmless.
 - If it's not allowed, we silently ignore the permission error.

This does not take a stance on whether these settings should be
enabled at all. This decision was already made when the table was
initially created. Astacus merely attempts to recreate what was
already there.

### DIFF
--- a/astacus/coordinator/plugins/clickhouse/client.py
+++ b/astacus/coordinator/plugins/clickhouse/client.py
@@ -24,7 +24,21 @@ class ClickHouseClient:
 
 
 class ClickHouseClientQueryError(Exception):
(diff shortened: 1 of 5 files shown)


aris-aiven (approved):
LGTM

PR #102, opened 2023-01-16
cassandra: allow backup/restore of a single datacenter

dmitry-potepalov:
In a multi-datacenter setup, you might want to have separate Astacus
clusters in different datacenter, each managing it's own part of the
cluster (e.g. because your datacenters mirror each other and you want
independent backups in different regions for disaster recovery).
If you configure the datacenter parameter in Cassandra plugin config,
Astacus will ignore nodes from other datacenters when creating a
backup and rewrite the keyspace replication settings accordingly on
backup and restore (setting NetworkTopologyStrategy to replicate
within the datacenter you specified and allowing you to restore a
single region when other datacenters are not available).

### DIFF
--- a/astacus/coordinator/plugins/cassandra/backup_steps.py
+++ b/astacus/coordinator/plugins/cassandra/backup_steps.py
@@ -9,11 +9,11 @@ from .model import CassandraConfigurationNode, CassandraManifest, CassandraManif
 from .utils import get_schema_hash
 from astacus.common import exceptions
 from astacus.common.cassandra.client import CassandraClient, CassandraSession
-from astacus.common.cassandra.schema import CassandraSchema
+from astacus.common.cassandra.schema import CassandraKeyspace, CassandraSchema
 from astacus.coordinator.cluster import Cluster
 from astacus.coordinator.plugins.base import Step, StepFailedError, StepsContext
 from dataclasses import dataclass
-from typing import Dict, List
+from typing import Dict, List, Optional
 
 import logging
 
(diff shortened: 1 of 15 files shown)


astacus/coordinator/plugins/cassandra/restore_steps.py
```diff
@@ -26,11 +28,21 @@
 logger = logging.getLogger(__name__)
 
 
+def _create_network_topology_strategy_cql(keyspace: CassandraKeyspace) -> str:
+    new_replication_strategy = cm.NetworkTopologyStrategy(keyspace.network_topology_strategy_dcs)
+    replication = new_replication_strategy.export_for_schema()
+    return f"CREATE KEYSPACE {keyspace.name} WITH replication = {replication} AND durable_writes = {keyspace.durable_writes}"
```
kmichel-aiven, line 34: Should we escape the keyspace name (because of how case is handled by Cassandra ?)

astacus/coordinator/plugins/cassandra/backup_steps.py
```diff
@@ -24,11 +24,28 @@ class CassandraSchemaRetrievalFailureException(exceptions.TransientException):
...
+
 def _retrieve_manifest_from_cassandra(
-    cas: CassandraSession, config_nodes: List[CassandraConfigurationNode]
+    cas: CassandraSession, config_nodes: List[CassandraConfigurationNode], *, datacenter: Optional[str]
 ) -> CassandraManifest:
     host_id_to_node: Dict[str, CassandraManifestNode] = {}
     for token, host in cas.cluster_metadata.token_map.token_to_host_owner.items():
+        if datacenter and host.datacenter != datacenter:
+            logger.debug(
+                "Skipping host %s, because it's datacenter %s doesn't match configured %s",
```
kmichel-aiven, line 43: nit. `its`

tests/unit/common/cassandra/test_schema.py
```diff
@@ -35,8 +35,9 @@ def test_schema(schema, mocker):
     )
 
     cks = schema.CassandraKeyspace(
+        network_topology_strategy_dcs={},
```
kmichel-aiven, line 38: We should have test where this new feature is exercised.

astacus/coordinator/plugins/cassandra/backup_steps.py
```diff
@@ -24,11 +24,28 @@ class CassandraSchemaRetrievalFailureException(exceptions.TransientException):
     pass
 
 
+def _remove_other_datacenters(keyspace: CassandraKeyspace, datacenter: str) -> None:
```
kmichel-aiven, line 27: This small helper and the other one should be tested.

kmichel-aiven (changes requested):

astacus/coordinator/plugins/cassandra/restore_steps.py
dmitry-potepalov: actually, I've got a better idea - let's ask the driver to do the magic for us

astacus/coordinator/plugins/cassandra/plugin.py
```diff
@@ -49,6 +49,7 @@ async def run_step(self, cluster: Cluster, context: StepsContext) -> None:
 class CassandraPlugin(CoordinatorPlugin):
     client: CassandraClientConfiguration
     nodes: Optional[List[CassandraConfigurationNode]]
+    datacenter: Optional[str]
```
kmichel-aiven, line 52: Sorry, I didn't see that earlier.
If astacus is updated and the config is not updated yet, will astacus fail to start because the value has no default value ?

(I'm thinking about intentional updates but also just a server or package manager pulling the last version by default)

dmitry-potepalov: according to [pydantic docs](https://docs.pydantic.dev/usage/models/#required-optional-fields), specifying Optional allows to omit the field, and `CoordinatorPlugin` is an `AstacusModel`. I don't think it's necessary to make this a "required optional" field (require saying `datacenter=None` in every configuration), this is an opt-in feature.

kmichel-aiven: I agree with you, it should not be required optional :+1: 

I disagree with the choice of syntax of pydantic though :disappointed:

kmichel-aiven (approved):

PR #103, opened 2023-01-25
Raise the level of some of the log messages

kmichel-aiven:
Running astacus with a DEBUG log level is very verbose, mainly
because of the underlying libraries for object storage.

But running it with INFO loses a lot of useful information about
backup progress.

Raise the level of some of astacus log messages to make it
more useful to run at INFO level.

Also adjust the log formatting to be able to locate the logging sources.


aris-aiven (approved):
That will be useful, thanks.

PR #104, opened 2023-01-30
cassandra: apply a sanity check to the client configuration path

aris-aiven:
The Cassandra client configuration path may be passed in the configuration but the corresponding file not exist at the time an operation is requested on the node.

Return an informative HTTP error instead of the generic fastAPI 500.


astacus/node/api.py
```diff
@@ -133,6 +133,16 @@ def cassandra(req: Union[ipc.NodeRequest, ipc.CassandraStartRequest], subop: ipc
...
+    if not n.config.cassandra:
+        raise HTTPException(status_code=500, detail="Cassandra node configuration not found")
+    config_path = n.config.cassandra.client.config_path
+    if not config_path:
+        raise HTTPException(status_code=500, detail="Expected Cassandra client configuration path value not found")
+    if not config_path.exists():
+        raise HTTPException(
+            status_code=500,
+            detail=f"Cassandra client configuration file not found at {config_path!r}",
+        )
```
fingon, line 145: This seems slightly unnecessary; the Path could be converted to `pydantic.FilePath` which ensures it exists. 

This is also somewhat unexpected place to check for it, though, as in theory some of the operations work without file too? (e.g. they require just ability to get client, which also works using host+port)?

aris-aiven: Converted the field to the `FilePath` type, removed the extra checks and added a test covering the failure scenario

fingon (approved):

PR #107, opened 2023-03-21
coordinator: expose busy concept

alanfranz:
In order to avoid an excessive system load when using both Astacus and
Cashew, we'd like to understand when Astacus is actually doing something
(can happen if prune is restarted).

### DIFF
--- a/astacus/coordinator/api.py
+++ b/astacus/coordinator/api.py
@@ -139,3 +139,8 @@ async def op_sub_result(*, op_name: OpName, op_id: int, c: Coordinator = Depends
     if not op.subresult_sleeper:
         return
     op.subresult_sleeper.wakeup()
+
+
+@router.get("/busy")
+async def is_busy(*, c: Coordinator = Depends()) -> bool:
+    return c.is_busy()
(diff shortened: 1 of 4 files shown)


fingon:
Doesn't new operation fail if there is already something running (with acquired lock?). So at worst case prune'll just cause some lock attempts which will fail?

Content itself LGTM.

alanfranz:
> Doesn't new operation fail if there is already something running (with acquired lock?). So at worst case prune'll just cause some lock attempts which will fail?
> 
> Content itself LGTM.

When doing the cashew-to-astacus migration, I'd really like the ability to query astacus, because I need to alternate between astacus and cashew in some situations. If I need to try doing a backup in order to see whether astacus is backing up anything it can become problematic; instead, I'd like to check who's backing up, wait if any of those two is doing anything, once they're both free of any task then I choose which system to use in order to perform the backup. cashew already supports listing in-progress backups, but astacus doesn't (if you don't know the URL, which is something that I'd rather not persist, it's just more state from my POV).

fingon (approved):

PR #109, opened 2023-03-28
Add option to remove leftover ClickHouse snapshots from disks

aris-aiven:
To clean up failed backups, each node is currently requested to remove leftover
snapshots from its shadow directories. Allow executing the ``SYSTEM UNFREEZE``
command that deletes frozen backup files from all disks at once. The command is 
run concurrently on all nodes.

### DIFF
--- a/tests/integration/coordinator/plugins/clickhouse/conftest.py
+++ b/tests/integration/coordinator/plugins/clickhouse/conftest.py
@@ -150,9 +150,14 @@ async def create_clickhouse_cluster(
 
 @contextlib.asynccontextmanager
 async def create_astacus_cluster(
-    storage_path: Path, zookeeper: Service, clickhouse_cluster: ClickHouseServiceCluster, ports: Ports
+    storage_path: Path,
+    zookeeper: Service,
+    clickhouse_cluster: ClickHouseServiceCluster,
+    ports: Ports,
+    *,
+    use_system_unfreeze: bool,
 ) -> AsyncIterator[ServiceCluster]:
-    configs = create_astacus_configs(zookeeper, clickhouse_cluster, ports, Path(storage_path))
+    configs = create_astacus_configs(zookeeper, clickhouse_cluster, ports, Path(storage_path), use_system_unfreeze)
     async with contextlib.AsyncExitStack() as stack:
         astacus_services_coro: List[Awaitable] = [stack.enter_async_context(_astacus(config=config)) for config in configs]
         astacus_services = list(await asyncio.gather(*astacus_services_coro))
@@ -209,6 +214,7 @@ def create_clickhouse_configs(
                     </zookeeper>
                     <mark_cache_size>5368709120</mark_cache_size>
                     <max_server_memory_usage_to_ram_ratio>0.5</max_server_memory_usage_to_ram_ratio>
+                    <enable_system_unfreeze>true</enable_system_unfreeze>
                     <user_directories>
                         <users_xml>
                             <path>{str(data_dir / "users.xml")}</path>
@@ -271,6 +277,7 @@ def create_astacus_configs(
     clickhouse_cluster: ClickHouseServiceCluster,
     ports: Ports,
     storage_path: Path,
+    use_system_unfreeze: bool,
 ) -> List[GlobalConfig]:
     storage_tmp_path = storage_path / "tmp"
     storage_tmp_path.mkdir(exist_ok=True)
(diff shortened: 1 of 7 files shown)


kmichel-aiven (dismissed):
SYSTEM UNFREEZE should be used for the `RemoveFrozenTablesStep`,
it's not necessary for the for the `UnfreezeTablesStep`.

It's useful for the `RemoveFrozenTablesStep` because when that step runs to cleanup previously failed backups, we don't know exactly what was the tables list at the time FREEZE was executed (plus the fact that it's much simpler to support multiple disks).

For the normal `UnfreezeTablesStep` I think it's better to keep the symmetry with the freeze, they both will run with the tables list collected from `RetrieveDatabasesAndTablesStep`.

aris-aiven:
Thanks for picking this up, was going through the docs and was over-eager to replace the `ALTER TABLE` with the simpler `SYSTEM` call. Probably also just needed a break haha

aris-aiven:
Does the SQL statement need to be executed on all nodes in a replicated context, or can we get away with executing it only once? If it only needs to be executed once, do we wish to keep the distributed operation timeout or is it superfluous?

kmichel-aiven:
> Does the SQL statement need to be executed on all nodes in a replicated context, or can we get away with executing it only once? If it only needs to be executed once, do we wish to keep the distributed operation timeout or is it superfluous?

It needs to be done once for each node, so the implementation will be similar to `SyncTableReplicasStep`.
So yes, that removes the need for the distributed operation timeout :+1: 
But keeping a query timeout with a large timeout would be good (so really similar to `SyncTableReplicasStep`, except for the loop on the tables and the `max_concurrent_sync` which isn't necessary since we don't loop on tables).

kmichel-aiven (changes requested):
Code looks good, but real tests with ClickHouse are missing.

You can extend the `params` of `fixture_restored_cluster` in `test_plugin.py`,
To make sure we test both with and without system unfreeze.

(And for now, run it locally, since ClickHouse is missing from CI.)

tests/integration/coordinator/plugins/clickhouse/conftest.py
```diff
@@ -38,7 +38,7 @@
 
 logger = logging.getLogger(__name__)
 
-pytestmark = [pytest.mark.clickhouse, pytest.mark.x86_64]
```
kmichel-aiven, line 41: Why removing the mark ?

aris-aiven: That was an oversight, was executing integration tests on arm, reverted that chunk. Happy to report that integration tests pass on both architectures.

aris-aiven:
When running integration tests pass on both x64 and arm architectures, I'm getting some errors on arm. Looking whether they're due to the changeset.

aris-aiven:
There is some flakiness on arm integration tests that is independent of this PR, and easily reproducible.

tests/integration/coordinator/plugins/clickhouse/test_plugin.py
```diff
@@ -43,15 +42,24 @@ def get_restore_steps_names() -> List[str]:
     return [step.__class__.__name__ for step in steps]
 
 
-@pytest.fixture(scope="module", name="restorable_cluster")
-async def fixture_restorable_cluster(ports: Ports, clickhouse_command: ClickHouseCommand) -> AsyncIterator[Path]:
+def _remove_frozen_tables_method(flag: bool) -> str:
+    return "SystemUnfreeze" if flag else "SnapshotClear"
+
+
+@pytest.fixture(scope="module", name="restorable_cluster", params=[True, False], ids=_remove_frozen_tables_method)
```
kmichel-aiven, line 49: Sorry, I missed it earlier:
It makes more sense to parametrize `use_system_unfreeze` on the restored cluster instead of the restorable cluster.
In the restorable cluster fixture, we never do a restore operation, we only do backup, so we don't exercise the unfreeze code.

kmichel-aiven (changes requested):

tests/integration/coordinator/plugins/clickhouse/test_plugin.py
aris-aiven: The restorable cluster takes the backup, so I set `use_system_unfreeze` in its config such that the flag is active during `RemoveFrozenTablesStep`. Am I missing something?
Did you rather mean to indirectly parametrize `fixture_restorable_cluster` using params in `fixture_restored_cluster` and build the test cases there?

kmichel-aiven: You didn't miss anything,  the test you wrote is completely fine.

I somehow mixed backup and restore while thinking about it :facepalm:

kmichel-aiven (approved):

PR #110, opened 2023-04-12
coordinator: do not keep all manifests in memory during cleanup

mhoffm-aiven:


kmichel-aiven (approved):

astacus/coordinator/cleanup.py
```diff
@@ -78,7 +78,12 @@ async def determine_kept_backups(self, *, retention, backups):
         if retention.minimum_backups is not None and retention.minimum_backups >= len(backups):
             return backups
         now = utils.now()
-        manifests = sorted(await self._download_backup_manifests(backups), key=lambda manifest: manifest.start, reverse=True)
+
+        aux = []
+        async for m in self._download_backup_manifests(backups):
+            aux.append((m.start, m.end, m.filename))
```
bridgetb, line 84: Out of interest, why is the second async for loop needed?

mhoffm-aiven: Discussed offline. To recap: having two doesn't hurt since amount of backups is limited and we don't download that much data.

PR #114, opened 2023-05-22
cassandra: revert client configuration path field type change

aris-aiven:
Astacus doesn't strictly need to depend on the Cassandra configuration file existing during startup. Backup management commands can be executed succesfully without the client configuration.

This is a partial revert of commit 8ba9255e472083aee643865143fca9d701cc4940


aris-aiven:
Superseded by #113

PR #115, opened 2023-05-29
 clickhouse: add support for multiple and remote disks 

kmichel-aiven:
The first thing is to support the disk prefix at the beginning of the file
path (anything before the store or shadow folder).

Some disks have metadata files instead of actual files in their store
folder, like object storage disks.

Even when the content is the same between replicas, the file
in object storage can be different, which makes the metadata files
different.

Because of that, we can't distribute parts among replicas and
we can't use a replicated ATTACH to exchange the data.

Even if we did, that wouldn't work great with all replication modes that
can be used in ClickHouse.

To fix that, stop doing parts distribution entirely and download all
parts from all nodes. To avoid the replication induced by ATTACH, replace
that with a SYSTEM RECOVER REPLICA: we remove the information in
ZooKeeper about the table replica, then force the disk content to be the
source of truth to recreate the ZooKeeper state.

When restoring a backup, we need to know if it's an old one that needs
to be restored with ATTACH/SYNC, or a new one that needs RESTART/RECOVER.
To detect that, introduce a version in the ClickHouse part of the manifest.

[DDB-498]

[DDB-498]: https://aiven.atlassian.net/browse/DDB-498?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -14,7 +14,7 @@ jobs:
       - name: Set up Python 3.10 with caching
         uses: actions/setup-python@v4
         with:
-          python-version: '3.10'
+          python-version: '3.11'
           cache: 'pip'
 
       - name: Install requirements
--- a/.github/workflows/tests-all.yml
+++ b/.github/workflows/tests-all.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: ["3.9", "3.10", "3.11"]
+        python-version: ["3.11"]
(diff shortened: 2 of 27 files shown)


aris-aiven:
Reviewing this tomorrow

aris-aiven (approved):
LGTM. Too bad the parts distribution optimization had to go, but happy the system restart & restore works well.

PR #116, opened 2023-06-06
clickhouse: extract the path of remote files 

kmichel-aiven:
After collecting metadata files of files in object storage, extract
the path of the remote files.

This required an extension of the snapshot protocol to have
different option for each glob.

### DIFF
--- a/astacus/common/ipc.py
+++ b/astacus/common/ipc.py
@@ -6,7 +6,7 @@ See LICENSE for details
 # pydantic validators are class methods in disguise
 # pylint: disable=no-self-argument
 
-from .magic import StrEnum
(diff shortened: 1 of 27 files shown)


kmichel-aiven:
To be rebased after #117 is merged.

alanfranz (approved):
LGTM, but I'm probably not the best person to understand the whole context and all the details. If nobody else is reviewing this, ping me next Monday and I'll try to do a more thorough review+merge.

aris-aiven:
Planning to do a review as well for sure

aris-aiven (approved):
LGTM as well

PR #117, opened 2023-06-06
 node: add type hints 

kmichel-aiven:
This is preparatory work for #116

Some changes related to `req` and `progress` are there to reduce the amount of `assert X is not None`.

### DIFF
--- a/astacus/node/snapshot.py
+++ b/astacus/node/snapshot.py
@@ -10,6 +10,7 @@ this module with proper parameters.
 
 """
 
+from ..common.rohmustorage import RohmuStorage
 from .node import NodeOp
 from .snapshotter import Snapshotter
 from .uploader import Uploader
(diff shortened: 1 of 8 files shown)


astacus/node/snapshot.py
```diff
@@ -10,6 +10,7 @@
 
 """
 
+from ..common.rohmustorage import RohmuStorage
```
aris-aiven, line 13: This should probably be an absolute import (not blocking)

aris-aiven (approved):
LGTM, thank you for taking the time to add these annotations. There's a single typo I think.

astacus/node/snapshot.py
kmichel-aiven: Good catch, I missed it in the PR split, the next one fixes it.

PR #118, opened 2023-06-08
storage: add type hints

kmichel-aiven:


alanfranz (approved):
tested with mypy, all ok.

PR #121, opened 2023-06-21
node: handle mode and owner of downloaded files

kmichel-aiven:
The goal is to be able to user a restored data dir when astacus and the restored service run
with a different user (but a shared group).

The first part is restoring less restricting file modes that are injected by `NamedTemporaryFile`
to enable read/write by the group on restored files.

The second part is more painful: despite having access to the restored files through the group
permission, some services also do file operations that require being the exact owner of the file.

This requires chowning the file, which cannot be done by the astacus or the service user.

This is solved by using sudo for the chown, and giving the sudo command a recognizable and
restricted shape that is suitable for sudoers directives.

### DIFF
--- a/astacus/node/download.py
+++ b/astacus/node/download.py
@@ -22,19 +22,24 @@ from typing import Callable, Dict, List, Optional, Sequence
 
 import base64
 import contextlib
+import getpass
 import logging
 import os
 import shutil
+import subprocess
 
 logger = logging.getLogger(__name__)
 
 
 class Downloader(ThreadLocalStorage):
-    def __init__(self, *, dst: Path, snapshotter: Snapshotter, parallel: int, storage: Storage) -> None:
+    def __init__(
+        self, *, dst: Path, snapshotter: Snapshotter, parallel: int, storage: Storage, copy_dst_owner: bool = False
+    ) -> None:
         super().__init__(storage=storage)
         self.dst = dst
         self.snapshotter = snapshotter
         self.parallel = parallel
+        self.copy_dst_owner = copy_dst_owner
 
     def _snapshotfile_already_exists(self, snapshotfile: ipc.SnapshotFile) -> bool:
         relative_path = snapshotfile.relative_path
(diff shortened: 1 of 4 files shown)


astacus/node/download.py
```diff
@@ -117,6 +124,29 @@ def _cb(*, map_in: Sequence[ipc.SnapshotFile], map_out: Sequence[ipc.SnapshotFil
...
+            for snapshotfile in snapshotstate.files:
+                absolute_path = self.dst / snapshotfile.relative_path
+                chown_candidate = absolute_path
+                while chown_candidate != self.dst and chown_candidate.stat().st_uid != dst_owner_uid:
+                    chowned_paths.add(str(chown_candidate.relative_to(self.dst)))
+                    chown_candidate = chown_candidate.parent
+            if chowned_paths:
+                astacus_user = getpass.getuser()
+                dst_owner_gid = self.dst.stat().st_gid
+                # We're very specific to allow a sufficiently restrictive sudoers configuration
```
aiven-amartin, line 140: Can we comment to link to where we set the sudoers? I do wonder if managing this just by adding the two users to the same group that we permission to the files is easier to follow, but as is I think this makes sense to me. We could also just explicitly pass the user in instead of unmasking the folder, since we know when instantiating the class.

aiven-amartin (commented):
Will poke more and probably approve later today, but did have the one question for now.

astacus/node/download.py
kmichel-aiven: We also share a group between astacus and the service that we backup, but that's not sufficient to avoid all permission errors when doing operations on the restored file (for instance, ClickHouse uses utime on some files and folders, which require being the exact owner, not merely having write access, hardlink operations are also quite restricted).

Where we set the sudoers is dependent on how astacus is deployed, aiven does it one way in core but it's not generalizable. (Or you just mean describing the typical content of a reasonable file in `/etc/sudoers.d/something` ?)

aiven-amartin: I was hoping the former was standardized, but otherwise I guess the latter?

aiven-amartin: No point relishing just for that though

aiven-amartin (approved):

PR #122, opened 2023-06-29
deps: update rohmu models for the latest version

kmichel-aiven:
The incomplete models and type mismatch were causing integration failures with the latest rohmu.


aris-aiven (approved):
Great, thanks for the fix

PR #123, opened 2023-07-11
zookeeper: add support for client authentication

aris-aiven:
Kazoo supports digest authentication and clients may want to authenticate their requests. Implement it for the ClickHouse plugin.

### DIFF
--- a/astacus/coordinator/plugins/zookeeper.py
+++ b/astacus/coordinator/plugins/zookeeper.py
@@ -10,6 +10,7 @@ from typing import AsyncIterator, Callable, Collection, Dict, List, Optional, Tu
 
 import asyncio
 import contextlib
+import dataclasses
 import enum
 import kazoo.exceptions
 import logging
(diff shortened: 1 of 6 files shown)


astacus/coordinator/plugins/zookeeper.py
```diff
@@ -124,6 +126,11 @@ def transaction(self) -> ZooKeeperTransaction:
         raise NotImplementedError
 
 
+class ZooKeeperUser(AstacusModel):
+    username: str
+    password: SecretStr
+
```
kmichel-aiven, line 132: I'd rather not have pydantic models go through all code, and restrict it to IO (config & ipc) → Having a model for config and a separate dataclass for the internals would be cleaner and avoid some future headache.

kmichel-aiven (changes requested):

aris-aiven:
Since we only use digest auth, the dependency on `pure-sasl` is not needed.

astacus/coordinator/plugins/zookeeper.py
aris-aiven: Makes sense, since we have to standard naming for the pydantic -> dataclass translation chose `to_dataclass`.

kmichel-aiven (approved):

PR #125, opened 2023-07-12
Implement basic partial restore for Cassandra

dmitry-potepalov:
Base it upon the node replacement for dead nodes (replace_address), but
skip the streaming stage, since we're downloading the data from object
storage. This is not production-ready yet, because the current process
loses all the writes since the last backup when restoring. Incremental
backup and restore to be introduced later.

After downloading the snapshot, stop Cassandra on the node we took the
snapshot of. That node might no longer be in the cluster, we attempt
to match it to configuration nodes the same way we match configuration
nodes to live cluster nodes when creating the backup.

Then start the replacement node with replace_address and
skip_bootstrap_streaming parameters. Wait until it is up, at this point
Cassandra should remove the old node from gossip because we instructed
it to perform a replacement.

### DIFF
--- a/astacus/coordinator/plugins/cassandra/plugin.py
+++ b/astacus/coordinator/plugins/cassandra/plugin.py
@@ -12,7 +12,14 @@ from astacus.common.cassandra.config import CassandraClientConfiguration, SNAPSH
 from astacus.common.snapshot import SnapshotGroup
 from astacus.coordinator.cluster import Cluster
 from astacus.coordinator.plugins import base
-from astacus.coordinator.plugins.base import CoordinatorPlugin, OperationContext, Step, StepFailedError, StepsContext
+from astacus.coordinator.plugins.base import (
+    CoordinatorPlugin,
+    MapNodesStep,
+    OperationContext,
+    Step,
+    StepFailedError,
+    StepsContext,
+)
 from astacus.coordinator.plugins.cassandra import backup_steps, restore_steps
 from dataclasses import dataclass
 from typing import List, Optional
@@ -95,6 +102,7 @@ class CassandraPlugin(CoordinatorPlugin):
             base.BackupNameStep(json_storage=context.json_storage, requested_name=req.name),
             base.BackupManifestStep(json_storage=context.json_storage),
             restore_steps.ParsePluginManifestStep(),
+            base.MapNodesStep(partial_restore_nodes=req.partial_restore_nodes),
         ] + cluster_restore_steps
 
     def get_restore_schema_from_snapshot_steps(self, *, context: OperationContext, req: ipc.RestoreRequest) -> List[Step]:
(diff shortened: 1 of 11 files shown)


tests/unit/coordinator/plugins/cassandra/test_restore_steps.py
```diff
@@ -42,20 +66,79 @@ async def test_step_start_cassandra(mocker, override_tokens):
...
+                return node_to_backup_index
+            case _:
+                raise NotImplementedError(cl)
+
+    run_subop = mocker.patch.object(restore_steps, "run_subop")
+
+    step = restore_steps.StopReplacedNodesStep(partial_restore_nodes=partial_restore_nodes, cassandra_nodes=cassandra_nodes)
+    context = SimpleNamespace(get_result=get_result)
+    cluster = SimpleNamespace(nodes=nodes)
+    result = await step.run_step(cluster, context)
```
kmichel-aiven, line 134: There are already similar instances in this file, but why mocking cluster and context ?
The real classes don't have dependencies and can be used in tests without having to
create a custom `get_result`:

```suggestion
    run_subop = mocker.patch.object(restore_steps, "run_subop")
    step = restore_steps.StopReplacedNodesStep(partial_restore_nodes=partial_restore_nodes, cassandra_nodes=cassandra_nodes)
    context = StepsContext()
    context.setResult(restore_steps.ParsePluginManifestStep, plugin_manifest)
    context.setResult(base.MapNodesStep, node_to_backup_index)
    cluster = Cluster(nodes=nodes)
    result = await step.run_step(cluster, context)
```

astacus/coordinator/plugins/cassandra/plugin.py
```diff
@@ -95,22 +114,27 @@ def get_restore_steps(self, *, context: OperationContext, req: ipc.RestoreReques
...
         ]
 
     def get_restore_schema_from_manifest_steps(self, *, context: OperationContext, req: ipc.RestoreRequest) -> List[Step]:
+        assert self.nodes is not None
         client = CassandraClient(self.client)
 
         return [
             # Start cassandra with backed up token distribution + set schema + stop it
-            restore_steps.StartCassandraStep(partial_restore_nodes=req.partial_restore_nodes, override_tokens=True),
+            restore_steps.StartCassandraStep(override_tokens=True, cassandra_nodes=self.nodes),
```
kmichel-aiven, line 137: nit. This is not urgent, but there are some benefits to not reusing the same class in the steps list, even when not using their result :
the class name is used as a metric name so monitoring about steps duration and failures can be confusing if two steps have the same name (and communication between operators).

It can be helpful to have small subclasses of `StartCassandraStep` that just pin the booleans values and have names that describe which start it is.

kmichel-aiven (approved):

PR #126, opened 2023-07-26
m3db: fix MapNodesStep order

kmichel-aiven:
This step needs to run before RewriteEtcdStep.


alanfranz (approved):
This problem was introduced in d450e92a3e9932ee5af1c6fb34272760c9d453ef - LGTM

PR #127, opened 2023-08-04
Retry failed attempts when reloading the config

kmichel-aiven:
If we try to reload the config while astacus is still starting, then the connection will immediately fail instead of waiting the full requests timeout.

This happens more frequently than expected when the config reload is managed by another automated system, if that system is fast enough.


kmichel-aiven:
> Seems to work, but I find the code and test quite complex (exponential_backoff does a sleep internally?) and tests are not verifying all behaviour (like, does it ultimately succeed and return True? does it respect the wait_completion time)? Could be some good new-developer ticket to expand on testing.

Good points, I've added the relevant tests here: https://github.com/aiven/astacus/pull/128

PR #130, opened 2023-08-11
Test that the URL table engine is restored correctly

aris-aiven:
We only need to test that the table structure is restored correctly along with the rest of the schema metadata

[DDB-198]

### DIFF
--- a/tests/integration/coordinator/plugins/clickhouse/test_plugin.py
+++ b/tests/integration/coordinator/plugins/clickhouse/test_plugin.py
@@ -17,7 +17,7 @@ from tests.integration.coordinator.plugins.clickhouse.conftest import (
     MinioBucket,
     run_astacus_command,
 )
-from typing import AsyncIterable, AsyncIterator, List, Sequence
+from typing import AsyncIterable, AsyncIterator, Final, List, Sequence
 from unittest import mock
 
 import base64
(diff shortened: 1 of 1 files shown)


tests/integration/coordinator/plugins/clickhouse/test_plugin.py
```diff
@@ -215,6 +218,13 @@ async def check_object_storage_data(cluster: List[ClickHouseClient]) -> None:
         assert response == expected_data
 
 
+@pytest.mark.asyncio
+async def test_restores_url_engine_table(restored_cluster: List[ClickHouseClient]) -> None:
+    for client in restored_cluster:
+        response = await client.execute(b"SELECT count() FROM system.tables WHERE name = 'url_engine_table'")
```
kmichel-aiven, line 224: Can you check the full `create_table_query` from system.tables instead of just the name ?

It's to make sure we didn't lose any part of the table definition.

kmichel-aiven (dismissed):

tests/integration/coordinator/plugins/clickhouse/test_plugin.py
aris-aiven: Done

kmichel-aiven (approved):

PR #131, opened 2023-08-14
Fix snapshotter not forgetting removed files

kmichel-aiven:
When using the `src==dst` mode of the snapshotter, the `relative_path_to_snapshotfile` mutable member keeps remembering files that don't exist anymore.

This variable is read after the snapshot and then used to select which files to upload. Since these files don't exist anymore, they can't be uploaded and the backup fails.


joelynch (approved):
Looks good

PR #132, opened 2023-08-14
node: The chown_paths may be too large to chown

RommelLayco:
The number of files to chown could be really large such that the argument list becomes to long.

### DIFF
--- a/astacus/node/download.py
+++ b/astacus/node/download.py
@@ -134,7 +134,10 @@ class Downloader(ThreadLocalStorage):
                 while chown_candidate != self.dst and chown_candidate.stat().st_uid != dst_owner_uid:
                     chowned_paths.add(str(chown_candidate.relative_to(self.dst)))
                     chown_candidate = chown_candidate.parent
-            if chowned_paths:
+
+            chunk_size = 1000
+            sorted_paths = sorted(chowned_paths)
+            for chunk_start in range(0, len(sorted_paths), chunk_size):
                 astacus_user = getpass.getuser()
                 dst_owner_gid = self.dst.stat().st_gid
                 # We're very specific to allow a sufficiently restrictive sudoers configuration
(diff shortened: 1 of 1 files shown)


astacus/node/download.py
```diff
@@ -144,7 +145,7 @@ def _cb(*, map_in: Sequence[ipc.SnapshotFile], map_out: Sequence[ipc.SnapshotFil
                     f"--from={astacus_user}:{dst_owner_gid}",
                     f"{dst_owner_uid}",
                     "--",
-                    *sorted(chowned_paths),
+                    *sorted(chowned_path),
```
reedwade, line 148: Probly need to drop the `sorted()`

astacus/node/download.py
```diff
@@ -134,7 +134,8 @@ def _cb(*, map_in: Sequence[ipc.SnapshotFile], map_out: Sequence[ipc.SnapshotFil
                 while chown_candidate != self.dst and chown_candidate.stat().st_uid != dst_owner_uid:
                     chowned_paths.add(str(chown_candidate.relative_to(self.dst)))
                     chown_candidate = chown_candidate.parent
-            if chowned_paths:
+
+            for chowned_path in chowned_paths:
                 astacus_user = getpass.getuser()
                 dst_owner_gid = self.dst.stat().st_gid
                 # We're very specific to allow a sufficiently restrictive sudoers configuration
```
kmichel-aiven, line 141: If the list is very long, we probably don't want to spawn thousand of subprocesses,
one for every single file, a chunked approach might be better:

```python
                # We're very specific to allow a sufficiently restrictive sudoers configuration
                chunk_size = 1000
                sorted_paths = sorted(chowned_paths)
                for chunk_start in range(0, len(sorted_paths), chunk_size):
                    cmd: Sequence[str | Path] = [
                        "/usr/bin/sudo",
                        "/usr/bin/chown",
                        f"--from={astacus_user}:{dst_owner_gid}",
                        f"{dst_owner_uid}",
                        "--",
                        *sorted_paths[chunk_start: chunk_start + chunk_size],
                    ]
                    subprocess.run(cmd, shell=False, check=True, cwd=self.dst)
```

kmichel-aiven (changes requested):
Formatting issue found by CI.

kmichel-aiven (approved):

PR #133, opened 2023-08-17
Update CodeQL analysis GitHub Action

aris-aiven:
Token permissions needed to be better specified

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -11,6 +11,11 @@
 #
 name: "CodeQL"
 
+permissions:
+  actions: read
+  contents: read
+  security-events: write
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -11,6 +11,9 @@
 #
 name: "CodeQL"
 
+permissions:
+  security-events: write
```
jlprat, line 15: Let's be completely explicit
```suggestion
permissions:
  actions: read
  contents: read
  security-events: write
```

aris-aiven: Sure, thanks

kmichel-aiven (approved):

PR #134, opened 2023-08-18
Bring Cassandra to a known state before restoring

dmitry-potepalov:
To avoid restore retries to fail in a loop, because the previous attempt left something in the data directory, stop Cassandra and clean the data directory before attempting restoration. To avoid accidentally deleting live node data, blacklist potentially damaging restore operations by default. Blacklist should be explicitly reset when configuring a node to perform restoration.

### DIFF
--- a/astacus/common/ipc.py
+++ b/astacus/common/ipc.py
@@ -195,6 +195,7 @@ class SnapshotClearRequest(NodeRequest):
 class CassandraSubOp(StrEnum):
     get_schema_hash = "get-schema-hash"
     remove_snapshot = "remove-snapshot"
+    remove_keyspaces = "remove-keyspaces"
     restore_snapshot = "restore-snapshot"
     restore_snapshot_with_schema = "restore-snapshot-with-schema"
     start_cassandra = "start-cassandra"
--- a/astacus/coordinator/plugins/cassandra/plugin.py
+++ b/astacus/coordinator/plugins/cassandra/plugin.py
@@ -115,6 +115,7 @@ class CassandraPlugin(CoordinatorPlugin):
             base.BackupManifestStep(json_storage=context.json_storage),
             restore_steps.ParsePluginManifestStep(),
             base.MapNodesStep(partial_restore_nodes=req.partial_restore_nodes),
+            CassandraRestoreSubOpStep(op=ipc.CassandraSubOp.stop_cassandra),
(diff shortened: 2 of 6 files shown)


astacus/common/ipc.py
```diff
@@ -195,13 +195,23 @@ class SnapshotClearRequest(NodeRequest):
...
     stop_cassandra = "stop-cassandra"
     take_snapshot = "take-snapshot"
 
 
+DESTRUCTIVE_CASSANDRA_SUBOPS = [
+    CassandraSubOp.remove_keyspaces,
+    CassandraSubOp.restore_snapshot,
+    CassandraSubOp.restore_snapshot_with_schema,
+    CassandraSubOp.start_cassandra,
+    CassandraSubOp.stop_cassandra,
```
kmichel-aiven, line 211: Isn't the default value too conservative ? The remove sounds dangerous enough, but having the restore/start/stop means that even the first restore on an empty cluster will fail with the default config ?

kmichel-aiven (changes requested):

astacus/common/ipc.py
dmitry-potepalov: well, even if we leave start/stop Cassandra out of it, it'll still fail, because the remove step will have to be executed anyways :) . And why should the first restore succeed with the default config? The idea is "prevent Astacus from damaging a running cluster if something calls /restore by mistake", and force-stopping all nodes is also quite a bad idea if they're handling client load. Restore ops can accidentally resurrect something deleted if we're restoring from an older backup (e.g. if tombstones expired and we're restoring an sstable containing the deleted rows).

kmichel-aiven: Ah right. So default astacus config can only do backup. And only if it has extra options it can do restore.
Since normally you don't do both at the same time and place that seem reasonable :+1: 

That being said, since operations are coupled anyway. Should we really expose the subops list as a config option ?
Maybe a "allow_destructive_operations" boolean is enough and more stable ?

(Stable in the sense that we tend to change the list of ops and subops and a user wouldn't really care about keeping up with that, they would just care about specifying if it's a backup-only node of a restorable node)

dmitry-potepalov: OK, replaced with an access level enum that allows to specify intent instead of explicitly listing subops.

kmichel-aiven (approved):

PR #135, opened 2023-08-27
clickhouse: copy object storage files on restore

kmichel-aiven:
When restoring from an object storage that is different from the default one,
copy files from the restoring storage config to the default storage config.

This adds support for restoring tables using object storage files
as well as tables using local storage.

This relies on an unmerged version of romhu to allow copying files
between buckets.

### DIFF
--- a/astacus/coordinator/plugins/clickhouse/async_object_storage.py
+++ b/astacus/coordinator/plugins/clickhouse/async_object_storage.py
@@ -4,10 +4,12 @@ See LICENSE for details
 """
 from abc import ABC, abstractmethod
 from pathlib import Path
+from rohmu import BaseTransfer
 from rohmu.errors import FileNotFoundFromStorageError
 from starlette.concurrency import run_in_threadpool
-from typing import Any, Mapping
+from typing import Any, Iterator, Mapping, Self, Sequence
 
+import contextlib
 import dataclasses
 import datetime
 import logging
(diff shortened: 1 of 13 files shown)


astacus/coordinator/plugins/clickhouse/async_object_storage.py
```diff
@@ -46,30 +57,70 @@ def delete_key(self, key: str) -> None:
...
         return [ObjectStorageItem(key=Path(item["name"]), last_modified=item["last_modified"]) for item in items]
 
     async def delete_item(self, key: Path) -> None:
         await run_in_threadpool(self.storage.delete_key, str(key))
 
+    async def copy_items_from(self, source: "AsyncObjectStorage", keys: Sequence[Path]) -> None:
+        if not isinstance(source, RohmuAsyncObjectStorage):
+            raise NotImplementedError("Copying items is only supported from another RohmuAsyncObjectStorage")
+        str_keys = [str(key) for key in keys]
+        await run_in_threadpool(self.storage.copy_items_from, source.storage, str_keys)
```
aris-aiven, line 93: What advantage is there in using `starlette`'s `run_in_threadpool` over stdlib's `loop.run_in_executor`?

aris-aiven (approved):
LGTM, integration tests passed locally

astacus/coordinator/plugins/clickhouse/async_object_storage.py
kmichel-aiven: I don't think it has a big impact for us but [starlette](https://github.com/encode/starlette/blob/7e6c936487bc6bc57f15679fcadb8b9589d43636/starlette/concurrency.py#L29-L35) uses anyio to use different implementations but we only use the vanilla asyncio & thread.
Though, they take care of contextvars, that's one less thing to worry about: https://github.com/agronholm/anyio/blob/a9fae31ccfef0803bf14ab65bd1e9b76c7fc4920/src/anyio/_backends/_asyncio.py#L2055-L2106

It's there mostly for consistency since we use the same function in other places of astacus.

PR #138, opened 2023-09-07
Implement delta backups for Cassandra, part 1

dmitry-potepalov:
Add the concept of a lighter "delta" backup to Astacus. Lighter "delta" backups contain only the data reported as such by the database itself (e.g. via commit logs or Cassandra incremental backups). Compared to the "regular" snapshot-based backups, delta backups:
* do not require listing the hexdigest storage to create the backup
* do not require to build a full list of files present in the database to determine the diff
* can be restored one on top of the other (during node replacement this allows to do "catch-up" on the replacement before draining&removing the old node, decreasing the window of time the cluster is running with RF-1 nodes)

The delta APIs re-use the download and snapshot methods already present, but:
* deltas are gathered and uploaded using a different Snapshotter instance targeted at the delta globs
* Cassandra incremental backups require the user to remove the incremental files once the user's done with them, so an additional Clear step is added at the end

This PR is also an invitation to a discussion: do we need a separate "delta backup" concept in Astacus? My main "pro" arguments are:
* enables the "catch-up" behavior during replacement
* looks more natural in a PITR context (technically one could increase backup frequency to "each 5 minutes" and include the commitlogs since the latest full backup to allow restoring to "somewhere in between", but that looks a bit counterintuitive: why would a snapshot include something that depends on a previous snapshot?)
* doesn't require to iterate through all the files (and hexdigests) every time we want to do a backup (might not be compelling enough on it's own, but it's a nice bonus)

### DIFF
--- a/astacus/client.py
+++ b/astacus/client.py
@@ -50,10 +50,6 @@ def _run_op(op, args, *, json=None, data=None) -> bool:
     return True
 
 
-def _run_backup(args) -> bool:
-    return _run_op("backup", args)
-
-
 def _run_restore(args) -> bool:
     json = {}
     if args.storage:
@@ -139,11 +135,11 @@ def print_list_result(result):
         print(tabulate(table, headers=headers, tablefmt="github"))
 
 
-def _run_list(args) -> bool:
+def _run_list(args, list_path: str) -> bool:
     storage_name = ""
     if args.storage:
         storage_name = f"?storage={args.storage}"
-    r = http_request(f"{args.url}/list{storage_name}", caller="client._run_list")
+    r = http_request(f"{args.url}/{list_path}{storage_name}", caller="client._run_list")
     if r is None:
         return False
     print_list_result(ipc.ListResponse.parse_obj(r))
@@ -189,6 +185,17 @@ def _check_reload_config(args) -> bool:
     return True
 
 
+def _add_backup(p_parent, url_path: str) -> None:
+    p_backup = p_parent.add_parser("backup", help="Request backup")
+    p_backup.set_defaults(func=lambda args: _run_op(url_path, args))
+
+
+def _add_list(p_parent, url_path: str) -> None:
+    p_delta_list = p_parent.add_parser("list", help="List backups")
+    p_delta_list.add_argument("--storage", help="Particular storage to list (default: all)")
+    p_delta_list.set_defaults(func=lambda args: _run_list(args, url_path))
+
+
 def create_client_parsers(parser, subparsers):
     default_url = f"http://localhost:{magic.ASTACUS_DEFAULT_PORT}"
     parser.add_argument("-u", "--url", type=str, help="Astacus REST endpoint URL", default=default_url)
@@ -210,8 +217,7 @@ def create_client_parsers(parser, subparsers):
         "--status", action="store_true", help="Returns a status code of 1 if the configuration needs to be reloaded"
     )
 
-    p_backup = subparsers.add_parser("backup", help="Request backup")
-    p_backup.set_defaults(func=_run_backup)
+    _add_backup(subparsers, "backup")
 
     p_restore = subparsers.add_parser("restore", help="Request backup restoration")
     p_restore.add_argument("--storage", help="Storage to use (default: configured)")
@@ -219,9 +225,7 @@ def create_client_parsers(parser, subparsers):
     p_restore.add_argument("--stop-after-step", type=str, help="Stop restore operation after the specified step")
     p_restore.set_defaults(func=_run_restore)
 
-    p_list = subparsers.add_parser("list", help="List backups")
-    p_list.add_argument("--storage", help="Particular storage to list (default: all)")
-    p_list.set_defaults(func=_run_list)
+    _add_list(subparsers, "list")
 
     p_cleanup = subparsers.add_parser("cleanup", help="Delete backups that should no longer be kept")
     p_cleanup.add_argument("--minimum-backups", type=int, help="Minimum number of backups to be kept")
@@ -235,3 +239,8 @@ def create_client_parsers(parser, subparsers):
     p_delete = subparsers.add_parser("delete", help="Delete specific old existing backup(s)")
     p_delete.add_argument("--backups", nargs="+", help="Backup names to be deleted")
     p_delete.set_defaults(func=_run_delete)
+
+    p_delta = subparsers.add_parser("delta", help="Delta backup commands")
+    sub_delta = p_delta.add_subparsers()
+    _add_backup(sub_delta, "delta/backup")
+    _add_list(sub_delta, "delta/list")
--- a/astacus/coordinator/api.py
+++ b/astacus/coordinator/api.py
@@ -4,8 +4,8 @@ See LICENSE for details
 """
 
 from .cleanup import CleanupOp
-from .coordinator import BackupOp, Coordinator, RestoreOp
-from .list import list_backups
+from .coordinator import BackupOp, Coordinator, DeltaBackupOp, RestoreOp
+from .list import list_backups, list_delta_backups
 from .lockops import LockOps
 from .state import CachedListResponse
 from astacus import config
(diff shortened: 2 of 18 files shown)


joelynch:
So is the idea that you would say once a day do a "full" backup (which is like the checkpoint) and then very frequently do the delta backups and then when restoring download the latest full backup and all the delta backups that occurred after that time?

I think this is great, makes sense to me.  The delta backups are quite different and I think trying to shoehorn them into the existing backups is probably not a good idea.

astacus/client.py
```diff
@@ -51,7 +51,7 @@ def _run_op(op, args, *, json=None, data=None) -> bool:
 
 
 def _run_backup(args) -> bool:
-    return _run_op("backup", args)
+    return _run_op("deltabackup" if args.delta else "backup", args)
```
kmichel-aiven, line 54: They are different enough that I think it would be better to have a separate command for it instead of an option. (For clarity, to keep the parallel with the coordinator API, and because if we add options to only backup or only delta backups, the argparse and help will be simpler and avoid "this option only applies if you also enable this other option").

astacus/client.py
```diff
@@ -143,7 +143,8 @@ def _run_list(args) -> bool:
     storage_name = ""
     if args.storage:
         storage_name = f"?storage={args.storage}"
-    r = http_request(f"{args.url}/list{storage_name}", caller="client._run_list")
+    method = "delta/list" if args.delta else "list"
```
kmichel-aiven, line 146: A bit confusing to say "method" for a path in an HTTP context :)

astacus/coordinator/api.py
```diff
@@ -81,6 +82,12 @@ async def backup(*, c: Coordinator = Depends(), op: BackupOp = Depends(BackupOp.
     return c.start_op(op_name=OpName.backup, op=op, fun=runner)
 
 
+@router.post("/deltabackup")
```
kmichel-aiven, line 85: For consistency with /delta/list and the same pattern used in node ops, this could be better as `/delta/backup`

astacus/coordinator/api.py
```diff
@@ -30,6 +30,7 @@ class OpName(StrEnum):
     """(Long-running) operations defined in this API (for coordinator)"""
 
     backup = "backup"
+    delta_backup = "deltabackup"
```
kmichel-aiven, line 33: Can this be kept consistent with the name and use `delta_backup` for the value or do we need the non-underscore ?

astacus/node/api.py
```diff
@@ -115,11 +130,26 @@ def upload_result(*, op_id: int, n: Node = Depends()):
...
+    op, _ = n.get_op_and_op_info(op_id=op_id, op_name=OpName.upload)
+    return op.result
+
+
 @router.post("/download")
 def download(req: ipc.SnapshotDownloadRequest, n: Node = Depends()):
     if not n.state.is_locked:
         raise HTTPException(status_code=409, detail="Not locked")
-    return DownloadOp(n=n, op_id=n.allocate_op_id(), stats=n.stats, req=req).start()
+    return DownloadOp(n=n, op_id=n.allocate_op_id(), stats=n.stats, req=req).start(n.get_or_create_snapshotter)
```
kmichel-aiven, line 152: This looks a bit too hard to follow with the init parameters but also a callback passed later, the init takes a `req` and then the callback will be called with things from `req` that was passed just before that. I think it's worth simplifying and just passing a configured snapshotter from the API to the Op, without callbacks.

kmichel-aiven (changes requested):
I'm also not sure the multiple snapshotter in the state and the clear operations align well with https://github.com/Aiven-Open/astacus/pull/139

astacus/coordinator/api.py
dmitry-potepalov: the routing and result routing was built with the assumption that the operation name is a single word. I cheated a bit and reused the "backup" op type, should be good enough.

dmitry-potepalov:
> I'm also not sure the multiple snapshotter in the state and the clear operations align well with https://github.com/Aiven-Open/astacus/pull/139 

took a look, I don't see any major issues, just introduce a delta Snapshot under a different key in app state and it should work. Could you maybe pinpoint some specific issue?

astacus/coordinator/plugins/base.py
```diff
@@ -152,10 +153,12 @@ class UploadBlocksStep(Step[List[ipc.SnapshotUploadResult]]):
 
     storage_name: str
     validate_file_hashes: bool = True
+    upload_request: str = "upload"
+    list_hexdigests_step: type = ListHexdigestsStep
```
kmichel-aiven, line 157: The type is too vague here (and I'm surprised mypy doesn't complain), it should be `type[Step[Set[str]]]`, this allows static validatoin that the return value of `context.get_result` has the correct type.




Ah, I know why mypy didn't complain :( 
```python
def build_node_index_datas(
    *, hexdigests, snapshots: List[ipc.SnapshotResult], node_indices: List[int]
) -> List[NodeIndexData]:
```

astacus/coordinator/plugins/base.py
```diff
@@ -164,7 +167,28 @@ async def run_step(self, cluster: Cluster, context: StepsContext) -> List[ipc.Sn
...
             node_index_datas,
             validate_file_hashes=self.validate_file_hashes,
+            upload_request=self.upload_request,
+        )
+
+
+@dataclasses.dataclass
+class SnapshotClearStep(Step[List[ipc.NodeResult]]):
+    """
+    Request to clear the source hierarcy of the snapshotter on all nodes.
```
kmichel-aiven, line 177: nit. hierarchy

astacus/node/api.py
```diff
@@ -173,3 +223,38 @@ def cassandra(req: Union[ipc.NodeRequest, ipc.CassandraStartRequest], subop: ipc
...
+            for group in req.groups
+            if not any(existing_group.root_glob == group.root_glob for existing_group in groups)
+        ]
+    return groups
+
+
+def snapshotter_from_snapshot_req(req: SnapshotReq, n: Node) -> Snapshotter:
+    groups = groups_from_snapshot_req(req)
+    snapshotter = n.get_or_create_snapshotter(groups)
+    assert snapshotter is not None
```
kmichel-aiven, line 252: Why do we need this assert ?

astacus/node/api.py
```diff
@@ -173,3 +223,38 @@ def cassandra(req: Union[ipc.NodeRequest, ipc.CassandraStartRequest], subop: ipc
...
+    return snapshotter
+
+
+def delta_snapshotter_from_snapshot_req(req: SnapshotReq, n: Node) -> Snapshotter:
+    groups = groups_from_snapshot_req(req)
+    snapshotter = n.get_or_create_delta_snapshotter(groups)
+    assert snapshotter is not None
```
kmichel-aiven, line 259: Same here, why do we need this assert ?

kmichel-aiven (dismissed):

astacus/coordinator/api.py
```diff
@@ -114,22 +120,30 @@ async def _list_backups(*, req: ipc.ListRequest = ipc.ListRequest(), c: Coordina
...
+
+
 @router.post("/cleanup")
 async def cleanup(*, op: CleanupOp = Depends(CleanupOp.create), c: Coordinator = Depends()):
     runner = await op.acquire_cluster_lock()
     return c.start_op(op_name=OpName.cleanup, op=op, fun=runner)
 
 
 @router.get("/{op_name}/{op_id}")
+@router.get("/delta/{op_name}/{op_id}")
```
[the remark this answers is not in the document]
dmitry-potepalov: because operation status url depends on the request path, therefore status urls for `/delta/X` look like `/delta/X/N`
```
    def start_op(self, *, op: Op, op_name: str, fun: Callable[[], Any]) -> Op.StartResult:
        ...
        url = self.request_url
        status_url = urlunsplit((url.scheme, url.netloc, f"{url.path}/{op.op_id}", "", ""))
```

astacus/coordinator/plugins/base.py
```diff
@@ -99,6 +97,7 @@ class SnapshotStep(Step[List[ipc.SnapshotResult]]):
     """
 
     snapshot_groups: Sequence[SnapshotGroup]
+    snapshot_request: str = "snapshot"
```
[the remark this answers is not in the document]
dmitry-potepalov: but this is `SnapshotStep`, not `SnapshotRequestV2`; steps are internal to the coordinator and not serialized/sent between nodes. The `snapshot_request` parameter of the step is either `snapshot` or `delta/snapshot` depending on the context.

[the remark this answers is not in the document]
kmichel-aiven: :facepalm: I misread the diff...

astacus/node/api.py
dmitry-potepalov: hmm, indeed seems that we don't; probably repeated the one from inside the op

kmichel-aiven (approved):

PR #139, opened 2023-09-11
snapshotter: add an SQLite backed snapshotter

joelynch:
This PR is a little long but the only interesting stuff is in

* astacus/node/sqlite_snapshot.py
* astacus/node/snapshotter.py
* astacus/node/snapshot.py

Splits the snapshotter class into `Snapshot` and `Snapshotter`.  This means that the `SnapshotGroups` can change during successive calls to the `/snapshot`, `/upload` and `/download` APIs and also simplifies the code a bit.

Adds new classes `SQLiteSnapshot` and `SQLiteSnapshotter`.  These are backed by a simple sqlite database and should use basically no memory.  These classes are enabled by the field `node.db_path`.

TODO:
* Fix tests
* Benchmarking to make sure that the new snapshotter is actually better

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -144,3 +144,5 @@ astacus/proto/*_pb2.py
 **/py.typed
 
 .idea
+
+.vscode/
--- a/astacus/common/progress.py
+++ b/astacus/common/progress.py
@@ -6,6 +6,8 @@ See LICENSE for details
 """
 
 from .utils import AstacusModel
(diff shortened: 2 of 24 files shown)


joelynch:
Some benchmarks:

Memory snapshotter:
memory


SQLite snapshotter:
sqlite

joelynch:
This is out of date, will fix up and reopen a PR.

PR #140, opened 2023-09-11
common: stream uploaded files

kmichel-aiven:
Instead of encrypting and compressing the entire file on disk and then uploading that.
Progressively compress, encrypt and upload at the same time.

This reduces the Disk IO load during backups and avoid filling the disk if there is not enough
space left and individual files are large (such as Cassandra SSTable or ClickHouse part files).

[DDB-543]


aris-aiven (approved):
LGTM 👍

PR #143, opened 2023-09-28
Adjust max concurrent queries for ClickHousePlugin

meatlink:


astacus/coordinator/plugins/clickhouse/steps.py
```diff
@@ -685,16 +686,27 @@ async def run_step(self, cluster: Cluster, context: StepsContext) -> None:
...
-                execute_with_timeout(
-                    client,
-                    self.attach_timeout,
-                    f"ALTER TABLE {table_identifier} ATTACH PART {escape_sql_string(part_name)}".encode(),
+                process_part(part_desc, client)
+                for part_descs in zip_longest(
+                    *[
+                        list_parts_to_attach(snapshot_result, self.disks, tables_by_uuid)
+                        for snapshot_result in backup_manifest.snapshot_results
+                    ]
```
kmichel-aiven, line 706: This change is good but I think it becomes hard to follow with the `zip_longest` and it does not fully reach the apparent goal of having at most `max_concurrent_attach/node_count` concurrent queries on each node.

Maybe something like that instead :
```python
        # Do all nodes at the same time
        await asyncio.gather(
            *[
                # But on each node, limit the number of concurrent attach
                gather_limited(
                    self.max_concurrent_attach_per_node,
                    [
                        execute_with_timeout(
                            client,
                            self.attach_timeout,
                            f"ALTER TABLE {table_identifier} ATTACH PART {escape_sql_string(part_name)}".encode(),
                        )
                        for table_identifier, part_name in
                        list_parts_to_attach(snapshot_result, self.disk_paths, tables_by_uuid)
                    ]
                )
                for client, snapshot_result in zip(self.clients, backup_manifest.snapshot_results)
            ]
        )
```

(In the context of Astacus, it's ok to create a new setting and deprecate/ignore the old one)

astacus/coordinator/plugins/clickhouse/plugin.py
```diff
@@ -83,6 +84,21 @@ class ClickHousePlugin(CoordinatorPlugin):
     max_concurrent_sync: int = 100
     use_system_unfreeze: bool = True
 
+    @root_validator(pre=True)
+    def _apply_max_concurrent_query_defaults(cls, values):  # pylint: disable=E0213
+        default_value = 10 * len(values["clickhouse"].nodes) if "clickhouse" in values else 100
```
kmichel-aiven, line 89: I'm not sure we should have complex conditional defaults, given that we can afford to deprecate&ignore old parameters, I'd rather just move to `_per_node`  concurrency limit for all limits.

kmichel-aiven (dismissed):

astacus/coordinator/plugins/clickhouse/plugin.py
meatlink: Thanks! I added _per_node and changed the code in steps - would it be better?

astacus/coordinator/plugins/clickhouse/steps.py
meatlink: Thanks! It looks like I was a bit too focused on transposing the loops and missed the whole point. Fixed.

astacus/coordinator/plugins/clickhouse/plugin.py
```diff
@@ -66,21 +66,21 @@ class ClickHousePlugin(CoordinatorPlugin):
...
+    max_concurrent_restore_replica_per_node: int = 10
     freeze_timeout: float = 3600.0
     unfreeze_timeout: float = 3600.0
     # Deprecated parameter, ignored
     attach_timeout: float = 300.0
-    max_concurrent_attach: int = 100
+    max_concurrent_attach_per_node: int = 10
     sync_tables_timeout: float = 3600.0
-    max_concurrent_sync: int = 100
+    max_concurrent_sync_per_node: int = 10
```
kmichel-aiven, line 83: All the old parameters should be kept (and ignored) to avoid failing if a newer astacus is accidentally installed in a place where the config still uses the old parameters.

I don't know how/why, but I misplaced the `# Deprecated parameter, ignored` comment, it should be just before `use_system_unfreeze` (all the attach/sync parameters are not deprecated). But you can move it in the right place and then put all the new deprecated options after that.

kmichel-aiven (changes requested):
Minor config issue but LGTM otherwise.

astacus/coordinator/plugins/clickhouse/plugin.py
meatlink: Thanks Kevin! Changed.
(again, sorry for the delay with the changes)

kmichel-aiven (approved):

astacus/coordinator/plugins/clickhouse/plugin.py
kmichel-aiven: Nothing to be sorry about :)

PR #144, opened 2023-09-28
Make sstable restoration configurable

dmitry-potepalov:
Instead of hard-coding sstable restoration glob and matching strategy, make it configurable from coordinator side. Allows to keep the subop enum small while allowing to add incremental backup restoration.
    
Presently we use 2 configurations:
* restore without system_schema, match tables by name
* restore with system_schema, match tables by id

Incremental backup restoration needs both these options, but with a different table glob (from /backups instead of /snapshots/astacus).

### DIFF
--- a/astacus/common/ipc.py
+++ b/astacus/common/ipc.py
@@ -195,9 +195,8 @@ class SnapshotClearRequest(NodeRequest):
 class CassandraSubOp(StrEnum):
     get_schema_hash = "get-schema-hash"
     remove_snapshot = "remove-snapshot"
-    unrestore_snapshot = "unrestore-snapshot"
-    restore_snapshot = "restore-snapshot"
-    restore_snapshot_with_schema = "restore-snapshot-with-schema"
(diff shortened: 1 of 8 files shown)


astacus/common/ipc.py
```diff
@@ -195,9 +195,8 @@ class SnapshotClearRequest(NodeRequest):
 class CassandraSubOp(StrEnum):
     get_schema_hash = "get-schema-hash"
     remove_snapshot = "remove-snapshot"
-    unrestore_snapshot = "unrestore-snapshot"
-    restore_snapshot = "restore-snapshot"
-    restore_snapshot_with_schema = "restore-snapshot-with-schema"
+    unrestore_sstables = "unrestore-sstables"
+    restore_sstables = "restore-sstables"
```
dmitry-potepalov, line 199: N.B.: since this is restore, I don't worry about backward compatibilty, since partial restore is not yet supported -> if we restore, all the nodes are likely to be on the newest version.

kmichel-aiven (approved):

PR #145, opened 2023-10-04
Allow extra fields in the config

joelynch:
Right now we cannot add extra fields to the config without it not being backwards compatible.  Should follow the rules of protobuf here where you can add new fields but they must be optional and Astacus should be able to cope with these fields missing.

This will be required for the SQLite work where a new field on the node config is being added.

Also since fastapi is available we can use `jsonable_encoder` to make `jsondict` a bit less awkward.

### DIFF
--- a/astacus/common/utils.py
+++ b/astacus/common/utils.py
@@ -14,6 +14,7 @@ from abc import ABC
 from contextlib import contextmanager
 from multiprocessing.dummy import Pool  # fastapi + fork = bad idea
 from pathlib import Path
+from fastapi.encoders import jsonable_encoder
 from pydantic import BaseModel
 from typing import Any, AsyncIterable, Callable, Dict, Final, Iterable, Optional, TypeVar, Union
 
@@ -21,7 +22,6 @@ import asyncio
 import datetime
 import httpcore
 import httpx
-import json as _json
 import logging
 import os
 import re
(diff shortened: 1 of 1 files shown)


fingon (changes requested):
Unfortunately this is not acceptable - we know that if we're using astacus version X, we need config X.

The unknown fields not being an error causes too much grief with typos. It was intentional design choice from the start to rather error early than allow for configuration which is meaningless.

If you add new fields, you can certainly make them optional to make them non-mandatory.

joelynch:
Okay I think in the config `allow_extra` might make sense if we want to prevent typos.

But right now with genbumps we run multiple Astacus versions in the same cluster and communication is happening with Pydantic models that do not allow additional fields.  This means we can _never_ add a field to the IPC models.

It is impossible for the following 3 to hold simultaneously.
1. Allow for smooth genbump where multiple versions are running in the same cluster.
2. Do not allow for additional fields when parsing the IPC requests.
3. Allow the protocol to ever evolve.

So I think that at least the IPC models (and probably any models in the coordinator API although I don't think there are any right now) we should reconsider this design choice.  There is prior art here in `gRPC` and other `RPC` frameworks allowing additional fields to be present.

I will put this in a different PR so since it is no longer about the config.

kmichel-aiven:
There is a system of [Features](https://github.com/Aiven-Open/astacus/commit/458c9e074557202da8ca8c9a4c360709acedaa53) between nodes: nodes can declare which features they support and coordinators can send a different IPC message if the related feature is supported.

It's a bit more tedious than ignoring optional fields, on the other hand this allows the coordinator to be more intentional when sending a message to older nodes: the non-optional values can be different if that helps, and the responses can also be handled differently.

(So far we've used features [twice](https://github.com/Aiven-Open/astacus/blob/d9a75285961f8cb852d483a533a6a0e14be2017c/astacus/node/api.py#L40-L44), we never did more than send a message with more parameters, but we can :slightly_smiling_face:)

PR #146, opened 2023-10-05
Support addressing_style for rohmu s3 configurations

bugant:
Support addressing_style for rohmu s3 configurations:  this can be useful to support S3-compatible object storage that requires to use the (deprecated?) path style addressing.


bugant:
This is not strictly needed at the moment, so we can probably think of resolving this by including models from rohmu directly?  If needed, will re-open it.

PR #148, opened 2023-10-09
Use Rohmu models directly

joelynch:
Some types were added to help with some debugging I was doing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -144,3 +144,5 @@ astacus/proto/*_pb2.py
 **/py.typed
 
 .idea
+
+.vscode/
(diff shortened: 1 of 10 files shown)


joelynch:
There might be some issues with this and `notifier` as discussed in the rohmu PR.

fingon (approved):
Otherwise this LGTM but I don't know about the clickhouse, and more specifically the string->bytes changes there. @kmichel-aiven ?

joelynch:
VSCode had red underlined them all but since Table is a pydantic model it converts str to bytes automatically so I didn't really need to make that change and can revert

kmichel-aiven (approved):
LGTM, the changes to bytes are correct too.

PR #149, opened 2023-10-16
Use latest rohmu version

joelynch:


Makefile
```diff
@@ -14,8 +14,8 @@ GENERATED = astacus/version.py $(GENERATED_PROTOBUFS)
 PYTHON = python3
 DNF_INSTALL = sudo dnf install -y
 
-ZOOKEEPER_VERSION = 3.8.2
```
joelynch, line 17: The mirror used to download here only has the latest zookeeper version.  I considered using the url

```
wget --no-verbose http://archive.apache.org/dist/zookeeper/-$(ZOOKEEPER_VERSION)/apache-zookeeper-$(ZOOKEEPER_VERSION)-bin.tar.gz
```

which has old versions, but this url is extremely slow.

fingon (approved):

PR #150, opened 2023-10-16
Release extra files in dst after upload

dmitry-potepalov:
Allow the coordinator to release the files that were uploaded during backup creation. Helps to free disk space after the backup was uploaded. Especially useful for Cassandra, since because of compactions we end up keeping around copies of old sstables.

### DIFF
--- a/astacus/coordinator/plugins/base.py
+++ b/astacus/coordinator/plugins/base.py
@@ -192,6 +192,36 @@ class SnapshotClearStep(Step[List[ipc.NodeResult]]):
 
 
 @dataclasses.dataclass
+class SnapshotReleaseStep(Step[List[ipc.NodeResult]]):
+    """
+    Request to release the files we don't need any more in the destination hierarchy.
+
+    Allows to free some disk space before the next backup happens.
+    """
+
+    async def run_step(self, cluster: Cluster, context: StepsContext) -> List[ipc.NodeResult]:
+        snapshot_results = context.get_result(SnapshotStep)
+        nodes_metadata = await get_nodes_metadata(cluster)
+        all_nodes_have_release_feature = nodes_metadata and all(
+            Features.release_snapshot_files.value in n.features for n in nodes_metadata
+        )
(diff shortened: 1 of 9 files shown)


astacus/node/snapshotter.py
```diff
@@ -273,3 +277,9 @@ def _result_cb(*, map_in: SnapshotFile, map_out: SnapshotFile) -> bool:
         progress.add_success()
 
         return changes
+
+    def release(self, hexdigest: str) -> None:
+        assert self.lock.locked()
+        assert self.src != self.dst
+        for snapshotfile in self.hexdigest_to_snapshotfiles.get(hexdigest, []):
```
joelynch, line 284: The snapshot will get out of sync with the disk here.  Is that is so that we don't recompute hashes in case a file is not deleted?  We will need to be careful to always re-link all files when creating a new snapshot in the SQLLite PR (if I ever get around to finishing that), also obviously if an upload is ever run after this step looks like it will skip the files.

joelynch: I don't think the upload will fail, just skip the missing files https://github.com/Aiven-Open/astacus/blob/master/astacus/node/uploader.py#L41.

I like the idea of marking released rows, I'll keep that in mind

astacus/coordinator/plugins/base.py
```diff
@@ -191,6 +191,36 @@ async def run_step(self, cluster: Cluster, context: StepsContext) -> List[ipc.No
...
+    """
+    Request to release the files we don't need any more in the destination hierarchy.
+
+    Allows to free some disk space before the next backup happens.
+    """
+
+    async def run_step(self, cluster: Cluster, context: StepsContext) -> List[ipc.NodeResult]:
+        snapshot_results = context.get_result(SnapshotStep)
+        nodes_metadata = await get_nodes_metadata(cluster)
+        all_nodes_have_release_feature = nodes_metadata and all(
```
joelynch, line 205: nit: `nodes_all_have_feature` could be a helper.

PR #151, opened 2023-10-17
Restore delta backups in Cassandra

dmitry-potepalov:
When doing a partial restore, apply all the incremental changes we've
uploaded when creating delta backups. After restoring the snapshot,
download and put sstables from incremental backups into Cassandra's data
directory.

### DIFF
--- a/astacus/common/ipc.py
+++ b/astacus/common/ipc.py
@@ -257,6 +257,7 @@ class CassandraRestoreSSTablesRequest(NodeRequest):
     table_glob: str
     keyspaces_to_skip: Sequence[str]
     match_tables_by: CassandraTableMatching
+    expect_empty_target: bool
 
 
 # coordinator.api
--- a/astacus/common/ipc.py
+++ b/astacus/common/ipc.py
@@ -9,11 +9,12 @@ See LICENSE for details
 from .magic import DEFAULT_EMBEDDED_FILE_SIZE, StrEnum
 from .progress import Progress
 from .utils import AstacusModel, now, SizeLimitedFile
+from astacus.common.snapshot import SnapshotGroup
 from datetime import datetime
 from enum import Enum
 from pathlib import Path
 from pydantic import Field, root_validator
-from typing import List, Optional, Sequence
+from typing import List, Optional, Sequence, Set
 
 import functools
 import socket
@@ -116,6 +117,26 @@ class SnapshotRequestV2(NodeRequest):
     groups: Sequence[SnapshotRequestGroup] = ()
     # Accept V1 request for backward compatibility if the controller is older
     root_globs: Sequence[str] = ()
+
+
+def create_snapshot_request(
+    snapshot_groups: Sequence[SnapshotGroup], *, node_features: Set[str]
+) -> SnapshotRequestV2 | SnapshotRequest:
+    if NodeFeatures.snapshot_groups.value in node_features:
+        return SnapshotRequestV2(
+            groups=[
+                SnapshotRequestGroup(
+                    root_glob=group.root_glob,
+                    excluded_names=group.excluded_names,
+                    embedded_file_size_max=group.embedded_file_size_max,
+                )
+                for group in snapshot_groups
+            ],
+        )
+    # This is a lossy backward compatibility since the extra options are not passed
+    return SnapshotRequest(
+        root_globs=[group.root_glob for group in snapshot_groups],
+    )
 
 
 class SnapshotHash(AstacusModel):
--- a/astacus/common/ipc.py
+++ b/astacus/common/ipc.py
@@ -10,6 +10,7 @@ from .magic import DEFAULT_EMBEDDED_FILE_SIZE, StrEnum
 from .progress import Progress
 from .utils import AstacusModel, now, SizeLimitedFile
 from datetime import datetime
+from enum import Enum
 from pathlib import Path
 from pydantic import Field, root_validator
 from typing import List, Optional, Sequence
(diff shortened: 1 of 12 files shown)


astacus/common/ipc.py
```diff
@@ -221,6 +221,7 @@ class CassandraRestoreSSTablesRequest(NodeRequest):
     table_glob: str
     keyspaces_to_skip: Sequence[str]
     match_tables_by: CassandraTableMatching
+    expect_empty_target: bool
```
kmichel-aiven, line 224: It looks like this message is also used in the normal restore case, not only the delta, won't that break compatibility of nodes of different versions ?

I guess that can be OK, since all nodes in a restore are supposed to be recent, unless the creation of a cluster perfectly overlaps with the deployment of a new astacus version ?

kmichel-aiven (commented):
Some question to double-check IPC but LGTM otherwise.

astacus/common/ipc.py
dmitry-potepalov: > unless the creation of a cluster perfectly overlaps with the deployment of a new astacus version ?
that'd have to be very perfect timing, and even if so - the cluster is not operational yet, we can simply drop the old nodes and deploy fresh ones

kmichel-aiven: Yes, that's quite unlikely, looks fine then.

astacus/coordinator/plugins/cassandra/restore_steps.py
```diff
@@ -77,6 +87,66 @@ def find_matching_cassandra_index(self, backup_node: CassandraManifestNode) -> O
...
 
 
+@dataclass
+class UploadFinalDeltaStep(Step[List[ipc.BackupManifest]]):
+    json_storage: AsyncJsonStorage
+    storage_name: str
+
+    async def run_step(self, cluster: Cluster, context: StepsContext) -> List[ipc.BackupManifest]:
+        replaced_nodes = context.get_result(StopReplacedNodesStep)
+        assert len(replaced_nodes) == 1, "Multiple node replacement not supported"
```
kmichel-aiven, line 97: This should probably be a `StepFailedError`, not an assert ?

astacus/coordinator/plugins/cassandra/restore_steps.py
```diff
@@ -77,6 +87,66 @@ def find_matching_cassandra_index(self, backup_node: CassandraManifestNode) -> O
...
+        assert snapshot_results[0].hashes is not None
+        upload_results = await self.upload_delta(replaced_node, snapshot_results[0].hashes, cluster=cluster)
+        return await self.upload_manifest(snapshot_results, upload_results, context=context)
+
+    async def snapshot_delta(
+        self, node: CoordinatorNode, *, cluster: Cluster, context: StepsContext
+    ) -> List[ipc.SnapshotResult]:
+        snapshot_results = await SnapshotStep(
+            snapshot_groups=delta_snapshot_groups(), snapshot_request="delta/snapshot", nodes_to_snapshot=[node]
+        ).run_step(cluster, context)
```
kmichel-aiven, line 111: I don't think step should call other steps.
Steps are not meant to be used as arbitrary function calls, they have impact on what is logged and how Astacus reports its progress for instance.
If you need to share code between steps, extract a function and use that function in the two steps.

astacus/coordinator/plugins/cassandra/plugin.py
```diff
@@ -162,12 +151,36 @@ def get_restore_schema_from_snapshot_steps(self, *, context: OperationContext, r
...
+            base.RestoreDeltasStep(
+                json_storage=context.json_storage,
+                storage_name=context.storage_name,
+                restore_delta_url=f"cassandra/{ipc.CassandraSubOp.restore_sstables}",
+                partial_restore_nodes=req.partial_restore_nodes,
+                restore_delta_request=restore_deltas_req,
+            ),
             restore_steps.StopReplacedNodesStep(partial_restore_nodes=req.partial_restore_nodes, cassandra_nodes=self.nodes),
+            restore_steps.UploadFinalDeltaStep(json_storage=context.json_storage, storage_name=context.storage_name),
+            base.RestoreDeltasStep(
```
kmichel-aiven, line 176: We should avoid directly reusing the same class multiple time in the steps list, that makes progress reporting more confusing. I'd rather see two named subclasses for `RestoreDeltasStep` that "configure" the base class, or just call non-step functions, and only pass truly dynamic parameters to these steps.

(This is also true for `CassandraRestoreSubOpStep`, and I don't think that should propagate further).

astacus/node/cassandra.py
```diff
@@ -175,6 +179,16 @@ def _match_table_by_name(self, keyspace_name: str, table_name_and_id: str) -> Di
...
 
+    def _ensure_target_is_empty(self, *, keyspace_name: str, table_path: Path) -> None:
+        # Ensure destination path is empty except for potential directories (e.g. backups/)
+        # This should never have anything - except for system_auth, it gets populated when we restore schema.
+        existing_files = [file_path for file_path in table_path.glob("*") if file_path.is_file()]
+        if keyspace_name == "system_auth":
+            for existing_file in existing_files:
+                existing_file.unlink()
+            existing_files = []
+        assert not existing_files, f"Files found in {table_path.name}: {existing_files}"
```
kmichel-aiven, line 190: This should probably be a `StepFailedError`, not an assert.

kmichel-aiven (changes requested):

astacus/coordinator/plugins/cassandra/restore_steps.py
dmitry-potepalov: IIRC the logging and progress are provided by whatever calls the step, but I agree, calling the step directly looks like a WTF. Extracted the necessary common logic.

kmichel-aiven (approved):

PR #152, opened 2023-10-26
snapshotter: Use an SQLite backed snapshotter

joelynch:
DO NOT MERGE still testing with aiven-core.

# Context

Astacus can currently take up reasonable amount of ram if there are many files, especially when performing a snapshot.  This can eat into available ram for the database itself.

# Changes

Split `Snapshotter` into 2 classes.  `Snapshot` responsible for holding all the data about a snapshot on disk, and `Snapshotter`, responsible for updating the snapshot to the current state.

Makes these classes generic and with 2 implementations, one holding all files in memory (equivalent to how Snapshotter currently works on master) and 1 with files stored on disk via SQLite.  The SQLite version uses constant memory and is not any slower in my benchmarks (in fact it may be a little faster).  You can look at this branch for benchmarks https://github.com/Aiven-Open/astacus/compare/joelynch/bench.

### DIFF
--- a/astacus/common/progress.py
+++ b/astacus/common/progress.py
@@ -6,6 +6,9 @@ See LICENSE for details
 """
 
 from .utils import AstacusModel
+from pyparsing import Iterable
+from typing import Optional
+from typing_extensions import Self, TypeVar
 
 import logging
 import math
@@ -15,7 +18,7 @@ logger = logging.getLogger(__name__)
 _log_1_1 = math.log(1.1)
 
 
-def increase_worth_reporting(value, new_value=None, *, total=None):
+def increase_worth_reporting(value: int, new_value: Optional[int] = None, *, total: int | None = None):
     """Make reporting sparser and sparser as values grow larger
     - report every 1.1**N or so
     - if we know total, report every percent
@@ -36,6 +39,9 @@ def increase_worth_reporting(value, new_value=None, *, total=None):
     return old_exp != new_exp
 
 
+T = TypeVar("T")
+
+
 class Progress(AstacusModel):
     """JSON-encodable progress meter of sorts"""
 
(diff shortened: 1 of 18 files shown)


astacus/node/sqlite_snapshot.py
```diff
@@ -0,0 +1,248 @@
...
+    def perform_snapshot(self, *, progress: Progress) -> None:
+        files = self._list_files_and_create_directories()
+        new_or_existing = self._compare_current_snapshot(files)
+        for_upsert = self._compare_with_src(new_or_existing)
+        with_digests = self._compute_digests(for_upsert)
+        self._upsert_files(with_digests)
+        self._con.execute("drop table if exists new_files;")
+        self._con.commit()
+
+    def _list_files_and_create_directories(self) -> Iterable[Path]:
```
fingon, line 110: This functionality feels odd to have duplicated (with somewhat different implementations) here and in memory snapshotter.

fingon (commented):
Partial review (few comments that I noticed when browsing through it). 

At least at some point memory peak usage was mostly pydantic models of the per-file snapshot results, when I played with `memray` so not sure if this improves scalability a lot. Your results seem different though which maybe just indicates I'm out of touch with what has happened recently or this helps in mostly smaller setups. Biggest memory bottleneck (at least in the controller on N-node larger setups) is that it deals with results from all nodes, at same time, using pretty memory inefficient Python data structures.

One thing to consider also is that do we really want to maintain two backends? It is extra complexity to deal with (just looking at this PR, for example), and at questionable benefit.

joelynch:
> At least at some point memory peak usage was mostly pydantic models of the per-file snapshot results, when I played with `memray` so not sure if this improves scalability a lot. Your results seem different though which maybe just indicates I'm out of touch with what has happened recently or this helps in mostly smaller setups. Biggest memory bottleneck (at least in the controller on N-node larger setups) is that it deals with results from all nodes, at same time, using pretty memory inefficient Python data structures.

My benchmarks are purely the snapshot / snapshotter itself.  A good step 2 here we could consider is breaking up the manifests into multiple files (at least a file per node) uploaded by the nodes themselves rather than the coordinator to deal with memory bottlenecks there.

> One thing to consider also is that do we really want to maintain two backends? It is extra complexity to deal with (just looking at this PR, for example), and at questionable benefit.

I have the 2 backends basically just for benchmarking, you're right there is no benefit to having both.  But being forced to write out common interfaces was good I think :) if we choose a new backend in the future this will be useful.

astacus/node/sqlite_snapshot.py
joelynch: Everything here is an iterator to not mess up the memory usage.  The original implementation adapted here would yield the same value twice if there are overlapping globs.  But the original implementation is probably also faster for very selective globs, since it might not need to recurse the whole tree (although I haven't checked this much).

fingon: Yeah, I don't really care which implementation is picked, just that there's one (as opposed to two) as I don't really see benefit in having both.

fingon (approved):

fingon:
I guess one thing to consider is also that if path is not set, we could just use sqlite with in-memory backend. Due to that, the whole memory backend seems pointless (although probably faster).

astacus/common/progress.py
```diff
@@ -6,6 +6,9 @@
 """
 
 from .utils import AstacusModel
+from pyparsing import Iterable
```
aris-aiven, line 9: Should be imported from `collections.abc`

PR #153, opened 2023-10-27
tests: add type checking and fix type bugs

joelynch:
You can still opt out of type checking by not typing a function.  But if types are specified, we should check that they are correct (according to mypy).  This makes it easier to run mypy on new code in `tests` to check that it is correct without having a lot of preexisting errors appear.


fingon (approved):
LGTM. We could consider transitioning from pylint to something else too; we have reduced its use elsewhere already.

PR #154, opened 2023-11-03
snapshotter: use wcmatch.glob.globmatch function

joelynch:
All existing python glob match functions equivalent to `match(path: Path, glob: str) -> bool` have serious issues.

1. `Path.match` does not accept `**` until python 3.13, see [here](https://github.com/python/cpython/issues/73435).  We need this for ClickHouse.
2. `fnmatch` accepts `*`, but this character is translated to regex `.*` and matches a `/` character, which is not what we want.

Neither of these functions passed the test I wrote.

To fix this mess, I've introduced a new 3rd party library `wcmatch` to handle the glob matching.  I did not want to have to use a 3rd party library, but the [matching logic is pretty complex](https://stackoverflow.com/questions/27726545/python-glob-but-against-a-list-of-strings-rather-than-the-filesystem?rq=3) and I did not want to write it myself.  In the (very distant) future when our minimum supported python is python3.13 we can strip this out.

### DIFF
--- a/astacus/common/progress.py
+++ b/astacus/common/progress.py
@@ -65,7 +65,6 @@ class Progress(AstacusModel):
         for item in i:
             yield item
             self.add_success()
-        self.done()
         return None
 
     def start(self, n) -> None:
--- a/astacus/node/memory_snapshot.py
+++ b/astacus/node/memory_snapshot.py
@@ -146,7 +146,7 @@ class MemorySnapshotter(Snapshotter[MemorySnapshot]):
             snapshotfile = self.snapshot.get_file(found_file.relative_path)
             if snapshotfile:
                 self.snapshot.remove_file(snapshotfile)
-            dst_path.unlink()
+            dst_path.unlink(missing_ok=True)
             if increase_worth_reporting(i):
                 logger.info("#%d. extra file: %r", i, found_file.relative_path)
             changes += 1
--- a/astacus/node/snapshot_op.py
+++ b/astacus/node/snapshot_op.py
@@ -101,3 +101,4 @@ class ReleaseOp(NodeOp[ipc.SnapshotReleaseRequest, ipc.NodeResult]):
         with self.snapshotter.lock:
             self.check_op_id()
             self.snapshotter.release(self.req.hexdigests, progress=self.result.progress)
+            self.result.progress.done()
--- a/astacus/node/sqlite_snapshot.py
+++ b/astacus/node/sqlite_snapshot.py
@@ -15,8 +15,11 @@ from pathlib import Path
 from typing import Iterable, Sequence
 from typing_extensions import override
 
+import logging
 import os
 import sqlite3
+
+logger = logging.getLogger(__name__)
 
 
 class SQLiteSnapshot(Snapshot):
@@ -106,15 +109,17 @@ class SQLiteSnapshotter(Snapshotter[SQLiteSnapshot]):
 
     def perform_snapshot(self, *, progress: Progress) -> None:
         files = self._list_files_and_create_directories()
-        new_or_existing = self._compare_current_snapshot(files)
-        for_upsert = self._compare_with_src(new_or_existing)
-        with_digests = self._compute_digests(for_upsert)
-        self._upsert_files(with_digests)
-        self._con.execute("drop table if exists new_files;")
-        self._con.commit()
+        with self._con:
+            self._con.execute("begin")
+            new_or_existing = self._compare_current_snapshot(files)
+            for_upsert = self._compare_with_src(new_or_existing)
+            with_digests = self._compute_digests(for_upsert)
+            self._upsert_files(with_digests)
+            self._con.execute("drop table if exists new_files;")
 
     def _list_files_and_create_directories(self) -> Iterable[Path]:
         """List all files, and create directories in src."""
+        logger.info("Listing files in %s", self._src)
         for dir_, _, files in os.walk(self._src):
             dir_path = Path(dir_)
             if any(parent.name == magic.ASTACUS_TMPDIR for parent in dir_path.parents):
@@ -156,9 +161,10 @@ class SQLiteSnapshotter(Snapshotter[SQLiteSnapshot]):
                 """
             )
             if not self._same_root_mode():
+                logger.info("Deleting files in %s that are not in current snapshot", self._dst)
                 for (relative_path,) in cur:
-                    os.unlink(self._dst / relative_path)
-            self._con.commit()
+                    assert isinstance(relative_path, str)
+                    (self._dst / relative_path).unlink(missing_ok=True)
             cur.execute(
                 """
                 insert into new_files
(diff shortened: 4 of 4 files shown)


kmichel-aiven:
If we have to add a dependency, then so be it :(

I'm worried about two things :
 - `wcmatch.glob.globmatch` functions does a lot of work to build its patterns every time it's called, and we call that for every file, using and storing https://facelessuser.github.io/wcmatch/glob/#translate before iterating on all files could be good.
 - Once we have that lib, we have the risk of relying on it (adding patterns only supported by that lib and not by Python 3.13).

I'm wondering if we could just replace the `f"shadow/{escaped_freeze_name}/store/**/*"` group of ClickHouse with two groups ?
 `f"shadow/{escaped_freeze_name}/store/*/*/*/*"` and `f"shadow/{escaped_freeze_name}/store/*/*/*/*/*"` and avoid the extra dependency.
The groups handle two paths:
- whatever/disk/store/3-first-char-of-table-uuid/table-uuid/part-name/file-name
- whatever/disk/store/3-first-char-of-table-uuid/table-uuid/projection-name/part-name/file-name

As far as I know, the only case where we have subfolder in frozen parts is if we have [projections](https://clickhouse.com/docs/en/sql-reference/statements/alter/projection) in the table:
https://github.com/ClickHouse/ClickHouse/blob/af469cecb6d5577e960446eb3766318ec83843de/src/Storages/MergeTree/DataPartStorageOnDiskFull.cpp#L32-L40

In any case, we should have some projection in `setup_cluster_content` and a related test to make sure they are restored. The bug you uncovered probably means they are lost during backup.

(m3db also uses `**` but at the beginning of the path, if I understand the bug report, that case is correctly supported by the existing code ?)

joelynch:
> Once we have that lib, we have the risk of relying on it (adding patterns only supported by that lib and not by Python 3.13).

Yes.  I have tried to quarantine its use to the `utils` package so that we do not rely on it too much.  This is a worry though.

> wcmatch.glob.globmatch functions does a lot of work to build its patterns every time it's called, and we call that for every file, using and storing [facelessuser.github.io/wcmatch/glob/#translate](https://facelessuser.github.io/wcmatch/glob/#translate) before iterating on all files could be good.

This is a good point, it is like 20x faster to compile it first and this shouldn't be too much of a change.  `path_matches_glob` here is not too much slower than `Path.match` though interestingly.

```python
In [6]: %timeit Path('hello/asdf/sdf/world').match('hello/*/*/world')
3.26 µs ± 34.8 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [7]: %timeit path_matches_glob('hello/asdf/sdf/world', 'hello/*/*/world')
3.63 µs ± 17.8 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [8]: %timeit path_matches_glob('hello/asdf/sdf/world', 'hello/**/world')
3.96 µs ± 13.1 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [13]: compiled = re.compile(wcmatch.glob.translate('hello/**/world')[0][0])

In [14]: %timeit compiled.match('hello/asdf/sdf/world') is not None
196 ns ± 1.26 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

> (m3db also uses ** but at the beginning of the path, if I understand the bug report, that case is correctly supported by the existing code ?)

`Path.match` is very unusual IMO and very different from bash globs which I think is the intuition everyone has for what the globs in the config file are supposed to mean.  I think the behavior of '*' at the prefix also acts the same as '**' which is really weird.  I think I would prefer not to use this function until it is fixed.

```python
In [15]: Path('abc/123').match('**')
Out[15]: True

In [16]: Path('abc/123').match('*')
Out[16]: True

In [20]: Path('abc/123/456').match('abc/**')
Out[20]: False

In [21]: Path('abc/123/456').match('abc/*')
Out[21]: False

In [23]: Path('abc/123/456').match('**/456')
Out[23]: True

In [24]: Path('abc/123/456').match('*/456')
Out[24]: True

In [25]: Path('abc/123/456/456').match('abc/*/456')
Out[25]: False
```

kmichel-aiven:
> > Once we have that lib, we have the risk of relying on it (adding patterns only supported by that lib and not by Python 3.13).
> 
> Yes. I have tried to quarantine its use to the `utils` package so that we do not rely on it too much. This is a worry though.
> 
> > wcmatch.glob.globmatch functions does a lot of work to build its patterns every time it's called, and we call that for every file, using and storing [facelessuser.github.io/wcmatch/glob/#translate](https://facelessuser.github.io/wcmatch/glob/#translate) before iterating on all files could be good.
> 
> This is a good point, it is like 20x faster to compile it first and this shouldn't be too much of a change. `path_matches_glob` here is not too much slower than `Path.match` though interestingly.
> 
> ```python
> In [6]: %timeit Path('hello/asdf/sdf/world').match('hello/*/*/world')
> 3.26 µs ± 34.8 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
> 
> In [7]: %timeit path_matches_glob('hello/asdf/sdf/world', 'hello/*/*/world')
> 3.63 µs ± 17.8 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
> 
> In [8]: %timeit path_matches_glob('hello/asdf/sdf/world', 'hello/**/world')
> 3.96 µs ± 13.1 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
> 
> In [13]: compiled = re.compile(wcmatch.glob.translate('hello/**/world')[0][0])
> 
> In [14]: %timeit compiled.match('hello/asdf/sdf/world') is not None
> 196 ns ± 1.26 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
> ```

:+1:

In addition to pure speed that is visible in microbenchmark, that will avoid thrashing memory with thousand of short lived objects.

> > (m3db also uses ** but at the beginning of the path, if I understand the bug report, that case is correctly supported by the existing code ?)
> 
> `Path.match` is very unusual IMO and very different from bash globs which I think is the intuition everyone has for what the globs in the config file are supposed to mean. The behavior of '*' at the prefix also acts the same as '**' which is really weird. I think I would prefer not to use this function until it is fixed.
> 
> ```python
> In [15]: Path('abc/123').match('**')
> Out[15]: True
> 
> In [16]: Path('abc/123').match('*')
> Out[16]: True
> 
> In [20]: Path('abc/123/456').match('abc/**')
> Out[20]: False
> 
> In [21]: Path('abc/123/456').match('abc/*')
> Out[21]: False
> 
> In [23]: Path('abc/123/456').match('**/456')
> Out[23]: True
> 
> In [24]: Path('abc/123/456').match('*/456')
> Out[24]: True
> 
> In [25]: Path('abc/123/456/456').match('abc/*/456')
> Out[25]: False
> ```

That's fair, let's use the library then.

joelynch:
Okay this one is ready now I think, I have put all the SnapshotGroup glob stuff into a little library `astacus/node/snapshot_groups.py`.

joelynch:
All green https://github.com/aiven/aiven-core/pull/40432

PR #155, opened 2023-11-07
Implement delta backup retention 

dmitry-potepalov:
Retain only the delta backups that are newer than the oldest base. Also make sure we do not release hashes referenced by delta manifests as dangling.


alanfranz (commented):
Some questions:

* is there a reason for having flags called retain_deltas in one step, delete_deltas in the other step, with opposite default behaviours? 

* is there a reason for having those flags and ifs at all? I have seen there're multiple services, and some (I suppose, I haven't checked the implementation and I don't remember, actually) don't support or don't need delta backups; in those cases, I'd suppose that methods like "deltas_to_delete" (and DownloadKeptDeltaManifestsStep) will just return an empty list, making the subsequent operation a no-op if delta deletion makes no sense.

dmitry-potepalov:
> is there a reason for having those flags and ifs at all?

refactored to avoid flags and ifs, I agree that it wasn't the best possible solution

> DeleteDanglingHexdigestsStep has no test that I can see

added one

joelynch (approved):
Looks good

PR #156, opened 2023-11-16
Put delta sqlite snapshot in a separate db

dmitry-potepalov:
Different snapshots should be independent from each other, there's no value in isolating them programmatically by e.g. using different tables.


joelynch (approved):
Amazing

PR #160, opened 2023-11-27
clickhouse: account for FREEZE not replicated in v23.8

joelynch:
See https://github.com/ClickHouse/ClickHouse/pull/52064

### DIFF
--- a/astacus/coordinator/plugins/clickhouse/plugin.py
+++ b/astacus/coordinator/plugins/clickhouse/plugin.py
@@ -17,6 +17,7 @@ from .steps import (
     CollectObjectStorageFilesStep,
     DeleteDanglingObjectStorageFilesStep,
     FreezeTablesStep,
+    GetVersionsStep,
     ListDatabaseReplicasStep,
     MoveFrozenPartsStep,
     PrepareClickHouseManifestStep,
@@ -53,7 +54,7 @@ from astacus.coordinator.plugins.base import (
 )
 from astacus.coordinator.plugins.zookeeper_config import ZooKeeperConfiguration
 from pathlib import Path
-from typing import List, Sequence
+from typing import Any, Sequence
 
 
 class ClickHousePlugin(CoordinatorPlugin):
@@ -94,12 +95,13 @@ class ClickHousePlugin(CoordinatorPlugin):
     max_concurrent_sync_per_node: int = 10
     use_system_unfreeze: bool = True
 
-    def get_backup_steps(self, *, context: OperationContext) -> List[Step]:
+    def get_backup_steps(self, *, context: OperationContext) -> Sequence[Step[Any]]:
         zookeeper_client = get_zookeeper_client(self.zookeeper)
         clickhouse_clients = get_clickhouse_clients(self.clickhouse)
         disks = Disks.from_disk_configs(self.disks)
         return [
             ValidateConfigStep(clickhouse=self.clickhouse),
+            GetVersionsStep(clickhouse_clients),
             # Cleanup old frozen parts from failed backup attempts
             RemoveFrozenTablesStep(
                 clients=clickhouse_clients,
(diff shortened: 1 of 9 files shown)


astacus/coordinator/plugins/clickhouse/steps.py
```diff
@@ -811,8 +821,46 @@ async def run_step(self, cluster: Cluster, context: StepsContext) -> None:
...
+
+
 async def execute_with_timeout(client: ClickHouseClient, timeout: float, query: bytes) -> None:
     # we use a session because we can't use the SETTINGS clause with all types of queries
     session_id = secrets.token_hex()
     await client.execute(f"SET receive_timeout={timeout}".encode(), session_id=session_id)
     await client.execute(query, session_id=session_id, timeout=timeout)
+
+
+CLICKHOUSE_VERSION_PARTITION_CMDS_NOT_REPLICATED = (23, 8)
```
kmichel-aiven, line 848: This seem oddly named, what we're really after is the version where ALTER ... FREEZE became unreplicated, not ALTER ... PARTITION.

astacus/coordinator/plugins/clickhouse/steps.py
```diff
@@ -811,8 +821,46 @@ async def run_step(self, cluster: Cluster, context: StepsContext) -> None:
...
+
+
+async def run_maybe_replicated_on_all_clients(
```
kmichel-aiven, line 851: This function is specific to the ALTER FREEZE problem, but it's named as if it was generic.

astacus/coordinator/plugins/clickhouse/plugin.py
```diff
@@ -154,8 +156,10 @@ def get_restore_steps(self, *, context: OperationContext, req: RestoreRequest) -
         clients = get_clickhouse_clients(self.clickhouse)
         disks = Disks.from_disk_configs(self.disks)
         source_disks = Disks.from_disk_configs(self.disks, storage_name=context.storage_name)
+        clickhouse_clients = get_clickhouse_clients(self.clickhouse)
         return [
             ValidateConfigStep(clickhouse=self.clickhouse),
+            GetVersionsStep(clickhouse_clients),
```
kmichel-aiven, line 162: I don't think you need that step for restore ?

tests/unit/coordinator/plugins/clickhouse/test_steps.py
```diff
@@ -1232,5 +1239,45 @@ async def test_delete_object_storage_files_step(tmp_path: Path) -> None:
...
+    assert result == [(23, 8), (23, 3)]
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "versions,called_node_indicies",
+    [
+        ([(23, 8), (23, 12)], [0, 1]),
+        ([(23, 3), (23, 8)], [0]),
+        ([(23, 3), (22, 8)], [0]),
```
kmichel-aiven, line 1260: Would be good to have a test where the single selected node is not the 0th.

alanfranz:
@joelynch codecov is saying the coverage of the PR is not great; do you agree or is it a false positive? if a false positive, please state so.

kmichel-aiven (changes requested):
Unrelated changes should be in separate commits (SQL formatting, list→sequence, etc..). Preferably after the core of the PR since we might need to EP this.

astacus/coordinator/plugins/clickhouse/steps.py
joelynch: I guess I was using the same language as here https://github.com/ClickHouse/ClickHouse/blob/master/src/Storages/PartitionCommands.h#L19.  Previously all of these were replicated except ATTACH, FETCH and DROP.  Now none are replicated.

This is a bit obscure and no one will understand this so I will change to specifically refer to FREEZE.

astacus/coordinator/plugins/clickhouse/steps.py
joelynch: See prev comment for reasoning but I will change.

joelynch:
> @joelynch codecov is saying the coverage of the PR is not great; do you agree or is it a false positive? if a false positive, please state so.

The message codecov posts is misleading.  This commit actually has very high if not perfect test coverage!  See here https://app.codecov.io/gh/Aiven-Open/astacus/commit/0fdb0905696d83ac5d00159ef9ca97020351f8b0

The message posted by codecov is actually comparing this commit to an ancient commit from ages ago, from which the repo now has lower coverage, but that is not this particular commits fault.

joelynch:
> Unrelated changes should be in separate commits (SQL formatting, list→sequence, etc..). Preferably after the core of the PR since we might need to EP this.

list->sequence was required here to get pylance and newer versions of mypy to accept the other changes, since the (previously) implicit return type of some of these methods was actually modified.  I can revert any sql formatting changes though.

astacus/coordinator/plugins/clickhouse/steps.py
kmichel-aiven: Makes sense :+1:

kmichel-aiven (approved):

PR #161, opened 2023-11-29
coordinator: Make cleanup keep only interesting manifests

fingon:
The endgame we really want is adding per-node manifests and some sort of summary manifest to be handled here, not keeping the 'too old' manifests in memory helps a lot in case of nodes which have already taken too many backups.

With this, we will load only up to the <amount kept+1> manifests to memory at most, so having e.g. 70 manifests in object storage and 5 kept means we at most have 6 in memory at a time.

(As seen in production.)

### DIFF
--- a/astacus/coordinator/plugins/base.py
+++ b/astacus/coordinator/plugins/base.py
@@ -533,6 +533,36 @@ class RestoreDeltasStep(Step[None]):
         await cluster.wait_successful_results(start_results=start_results, result_class=ipc.NodeResult)
 
 
+def _prune_manifests(manifests: List[ipc.BackupManifest], retention: Retention) -> List[ipc.BackupManifest]:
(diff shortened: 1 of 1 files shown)


astacus/coordinator/plugins/base.py
```diff
@@ -533,6 +533,36 @@ async def clear_delta(
         await cluster.wait_successful_results(start_results=start_results, result_class=ipc.NodeResult)
 
 
+def _prune_manifests(manifests: List[ipc.BackupManifest], retention: Retention) -> List[ipc.BackupManifest]:
+    manifests = sorted(manifests, key=lambda m: (m.start, m.end, m.filename), reverse=True)
```
fingon, line 537: NB: Not sure if some later step makes assumptions about order -> sort even if we don't change the list here.

kmichel-aiven (approved):

PR #164, opened 2023-11-30
ignore tags file

sjamgade:


.gitignore
```diff
@@ -146,3 +146,4 @@ astacus/proto/*_pb2.py
 .idea
 
 .vscode/
+tags
```
kmichel-aiven, line 149: What are these tags file ?

sjamgade: genenrated by ctags

kmichel-aiven (approved):

PR #165, opened 2023-11-30
tests: Add combined coverage handling

fingon:
This slows invoked subprocesses by ~30% (or less, if they do heavy computation) which is hopefully acceptable cost. We already did gather coverage on the main pytest process.

I wonder if this breaks codecov or if it is somehow broken anyway - https://app.codecov.io/gh/Aiven-Open/astacus/pull/165 shows this as in 'processing' state for half an hour now. Perhaps we're not classified as open source plan and are hitting some limit? Or different coverage invocation makes it unable to find the parent commit (it has coverage % which is higher than what is in `main`).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,14 +55,17 @@ copyright:
 	$(eval MISSING_COPYRIGHT := $(shell git ls-files "*.py" | grep -v __init__.py | xargs grep -EL "Copyright \(c\) 20.* Aiven|Aiven license OK"))
 	@if [ "$(MISSING_COPYRIGHT)" != "" ]; then echo Missing Copyright statement in files: $(MISSING_COPYRIGHT) ; false; fi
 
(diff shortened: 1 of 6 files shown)


kmichel-aiven:
Do you plan to fix the 'processing' state first or the PR is already useful in the current state ?

(I tried to check https://app.codecov.io/gh/Aiven-Open/astacus/pull/165 but it says "Missing base commit" and "Coverage data is unknown", not sure if it's because the recent fixes to the base branch were not yet in place or something else.)

fingon:
TBH I don't really know what is wrong - I think the options are either merge it and hope (and revert if it breaks the coverage in general) or not do this at all.

kmichel-aiven:
> TBH I don't really know what is wrong - I think the options are either merge it and hope (and revert if it breaks the coverage in general) or not do this at all.

Ok, that's not too hard to revert in case of problems.

PR #166, opened 2023-12-04
clickhouse: fix slow list of tables

kmichel-aiven:
When listing all tables, we're only interested in the tables inside the Replicated database engine. In addition to that, we don't want ClickHouse to scan other database engines, some can be very slow to scan, like the ones querying remote databases (PostgreSQL, MySQL...).

ClickHouse is able to avoid scanning the tables of unwanted engines if
 the query is simple enough for it to recognize which databases are
 interesting. But the query we used did not allow that.

 Split the query in two to ensure this is the case.

 [DDB-777]


aris-aiven (approved):
Very nice, thank you

PR #168, opened 2023-12-05
Remove memory snapshot and snapshotter

dmitry-potepalov:
We don't want to support memory-based snapshot, because it consumes too much memory when there's a lot of files in the snapshot.


joelynch:
> it loses state on restart

The SQLite snapshotter currently always creates a new db on restart so does not have this benefit.  But I think if you have a careful look at the code, everything that needs to be is in a transaction so we could have it reuse an existing db.  Preferably this would be tested somehow though.

tests/integration/coordinator/plugins/clickhouse/conftest.py
```diff
@@ -564,6 +564,7 @@ def create_astacus_configs(
             ),
             node=NodeConfig(
                 root=clickhouse_service.data_dir,
+                db_path=storage_path,
```
kmichel-aiven, line 567: This breaks the tests, because then all nodes share the same folder for their db, and when they snapshot at the same time they open and try to overwrite the same files (which sqlite blocks, but it still fails).

Doing something like that fixes the tests:
```python
    node_db_paths = [storage_path / f"db_{node_port}" for node_port in node_ports]
    for node_db_path in node_db_paths:
        node_db_path.mkdir(exist_ok=True)
```

kmichel-aiven:
> > it loses state on restart
> 
> The SQLite snapshotter currently always creates a new db on restart so does not have this benefit. But I think if you have a careful look at the code, everything that needs to be is in a transaction so we could have it reuse an existing db. Preferably this would be tested somehow though.

By the way, this should be changed:
https://github.com/Aiven-Open/astacus/blob/17eb12524470a071a768b3aa69775715f19271a1/astacus/node/sqlite_snapshot.py#L26-L51

Implicit removal of files just by creating an object is dangerously surprising.
`__init__` should only connect objects together, not do things (especially things that have an impact on disk).

dmitry-potepalov:
> The SQLite snapshotter currently always creates a new db on restart so does not have this benefit

yes, seems I've been too optimistic :) amended the commit message and PR description

kmichel-aiven (approved):

PR #169, opened 2023-12-05
clickhouse: add function table test case

kmichel-aiven:


joelynch (approved):
No one has ever said clickhouse doesn't have enough features.

PR #170, opened 2023-12-11
cleanup: Use less memory

joelynch:
* `JsonStorage`: allow downloading a JSON to file rather than memory.
* Add a method to `Step` that allows a cleanup function to run for that step
* `BackupManifestFileWrapper`: wraps a manifest written to disk.

Now we can keep manifests on disk and only use when needed, rather than holding all in memory at the same time.

### DIFF
--- a/astacus/common/asyncstorage.py
+++ b/astacus/common/asyncstorage.py
@@ -7,6 +7,7 @@ See LICENSE for details
 
 from astacus.common.storage import HexDigestStorage, Json, JsonStorage
 from astacus.common.utils import AstacusModel
+from pathlib import Path
 from starlette.concurrency import run_in_threadpool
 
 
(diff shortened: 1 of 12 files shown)


fingon (changes requested):
This looks overly complex. What we already did was cache jsons locally (in the caching layer which is always used in practise), and this adds the cleanup steps which are basically just extra complexity. Not keeping the manifests in memory was trivial - simply just read them one by one and use them, and re-read in the rare case you need to do it again (coming from cache, it is fast enough, ujson load step).

Similarly I'm not that convinced BackupManifestFileWrapper is really that much less effort than splitting to the per-node manifests (after which the main manifest is only metadata, which coincidentally looks to some extent like what BackupManifestFileWrapper has).

joelynch:
Going to have another go at this.  Will do per node manifests in a different PR most likely and will make this one much simpler.

PR #172, opened 2023-12-15
clickhouse: fix restore with allow_nullable_key

kmichel-aiven:
If the setting was transiently set when creating a table, we need to support it when restoring the same table.

Like the other similar settings, it's harmless to always enable it, we're just accepting a table that was already accepted before the backup was done.

[DDB-791]


alanfranz (approved):
LGTM

PR #173, opened 2023-12-20
cleanup: reduce memory usage

joelynch:
Minimal fix for large memory usage during cleanup.  Rather than store all manifests in memory, reread them one at a time.

### DIFF
--- a/astacus/coordinator/plugins/base.py
+++ b/astacus/coordinator/plugins/base.py
@@ -49,7 +49,10 @@ class CoordinatorPlugin(AstacusModel):
                 explicit_delete=explicit_delete,
             ),
             DeleteBackupManifestsStep(json_storage=context.json_storage),
-            DeleteDanglingHexdigestsStep(hexdigest_storage=context.hexdigest_storage),
+            DeleteDanglingHexdigestsStep(
+                json_storage=context.json_storage,
+                hexdigest_storage=context.hexdigest_storage,
+            ),
         ]
 
 
@@ -533,7 +536,18 @@ class RestoreDeltasStep(Step[None]):
         await cluster.wait_successful_results(start_results=start_results, result_class=ipc.NodeResult)
 
 
-def _prune_manifests(manifests: List[ipc.BackupManifest], retention: Retention) -> List[ipc.BackupManifest]:
+@dataclasses.dataclass
+class ManifestMin:
+    start: datetime.datetime
+    end: datetime.datetime
+    filename: str
+
+    @classmethod
+    def from_manifest(cls, manifest: ipc.BackupManifest) -> ManifestMin:
+        return cls(start=manifest.start, end=manifest.end, filename=manifest.filename)
+
+
+def _prune_manifests(manifests: List[ManifestMin], retention: Retention) -> List[ManifestMin]:
     manifests = sorted(manifests, key=lambda m: (m.start, m.end, m.filename), reverse=True)
     if retention.minimum_backups is not None and retention.minimum_backups >= len(manifests):
         return manifests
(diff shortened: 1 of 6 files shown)


astacus/coordinator/plugins/base.py
```diff
@@ -575,32 +595,29 @@ class ComputeKeptBackupsStep(Step[Sequence[ipc.BackupManifest]]):
...
         return kept_manifests
 
-    async def compute_kept_basebackups(self, context: StepsContext) -> List[ipc.BackupManifest]:
+    async def compute_kept_basebackups(self, context: StepsContext) -> List[ManifestMin]:
         all_backup_names = context.get_result(ListBackupsStep)
         kept_backup_names = all_backup_names.difference(set(self.explicit_delete))
         manifests = []
         for backup_name in kept_backup_names:
-            manifests.append(await download_backup_manifest(self.json_storage, backup_name))
-            manifests = _prune_manifests(manifests, self.retention)
```
joelynch, line 590: Was this intended to run this repeatedly?  Looks wrong.

astacus/coordinator/plugins/base.py
```diff
@@ -533,7 +536,24 @@ async def clear_delta(
         await cluster.wait_successful_results(start_results=start_results, result_class=ipc.NodeResult)
 
 
-def _prune_manifests(manifests: List[ipc.BackupManifest], retention: Retention) -> List[ipc.BackupManifest]:
+@dataclasses.dataclass
+class ManifestMin:
+    start: datetime.datetime
+    end: datetime.datetime
+    filename: str
+    plugin_data: Dict[str, Any]
```
kmichel-aiven, line 544: This can be huge for ClickHouse, since it contains the path of all object storage items on remote disk, it shouldn't be part of the `ManifestMin`.

astacus/coordinator/plugins/base.py
fingon: It was needed for full manifests as we didn't want to download them all and only then start pruning them.

tests/unit/coordinator/plugins/test_base.py
```diff
@@ -118,7 +119,14 @@ def named_manifest(filename: str) -> ipc.BackupManifest:
     return make_manifest(some_isoformat, some_isoformat, filename=filename)
 
 
+# Global used but just for test case
+# pylint: disable=global-statement
+COUNT = 0
```
kmichel-aiven, line 124: No mutable globals, even just for tests.

kmichel-aiven (changes requested):

fingon (commented):
This isn't really what per node manifests are though. The point would be to split the Manifest so that per-node data is in separate sub-model which is referred to by e.g. name (so none of the data needs to be downloaded unless it is needed and even then the atomic unit is data for single node of single backup, as opposed to the whole backup). 

How that would work:
- when making backup, each node returns just reference to name of its sub-manifest (which it would upload on its own), instead of the data, to the coordinator
- coordinator gathers list of these (+ metadata) to Manifest and sticks it to object storage (quite small json file)

When cleaning up, per node manifests can must be traversed to identify the data in use, but even then it is only data for single node (as opposed to all nodes in cluster) that needs to be downloaded or be in memory at same time.

tests/unit/coordinator/plugins/test_base.py
joelynch: I could use a uuid instead?

kmichel-aiven: the functions is called twice, maybe just a parameter that is either 1 or 2 would be enough ?

kmichel-aiven (approved):

PR #176, opened 2024-01-17
lower log level of GET calls to DEBUG

kmichel-aiven:
`POST` and `PUT` calls are much more interesting than `GET` calls, at  least from the point of view of the server.

Log `GET` requests at the `DEBUG` level while logging other methods at the `INFO` level.

### DIFF
--- a/astacus/common/access_log.py
+++ b/astacus/common/access_log.py
@@ -1,0 +1,16 @@
+"""
+Copyright (c) 2024 Aiven Ltd
+See LICENSE for details
+"""
+from collections.abc import Sequence
+
+import logging
+
+
+class AccessLogLevelFilter(logging.Filter):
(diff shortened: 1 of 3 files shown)


astacus/common/access_log.py
```diff
@@ -0,0 +1,16 @@
...
+"""
+from collections.abc import Sequence
+
+import logging
+
+
+class AccessLogLevelFilter(logging.Filter):
+    def filter(self, record: logging.LogRecord) -> bool:
+        # Lower the log level of GET requests from INFO to DEBUG
+        if isinstance(record.args, Sequence) and len(record.args) >= 2 and record.args[1] == "GET":
```
alanfranz, line 13: do we need to do all this work at each request? Isn't it possible (or desiderable) to do this at astacus level with a logger.debug vs logger.info call in the right place?

PR #177, opened 2024-01-18
filter access logs by levels

kmichel-aiven:
The previous configuration was checking the level at the logger step bot not at the handler step, so logs were not filtered according to their modified log level.


alanfranz (approved):
I'm unable to double-check, but it cannot break things, worst case there'll be another PR.

PR #178, opened 2024-01-31
reduce astacus memory usage [DDB-853]

kmichel-aiven:
This is not the optimal solution (splitting files, changing the manifest format, etc.)

But profiling shows significant reduction in memory usage for large manifests with these changes, without breaking backward compatibility:

- use `str` instead of `Path` wherever possible. The `Path` API is nice, but  the abstraction is too expensive when we have a large number of objects.
- only convert to `pydantic` models the subset of data that we care about, for instance if we want a single snapshot in an entire manifest, don't convert the entire manifest, just the single snapshot.
- `list_backups`: stay completely in json-land, but with type-safe accessors to read the data.

Tested on a large 800MiB manifest, peak usage during `list_backups` was reduced from ~8GiB to ~4.1GiB and lasted for a much shorter time.

### DIFF
--- a/astacus/common/ipc.py
+++ b/astacus/common/ipc.py
@@ -74,7 +74,7 @@ class MetadataResult(AstacusModel):
 
 @functools.total_ordering
 class SnapshotFile(AstacusModel):
-    relative_path: Path
+    relative_path: str
     file_size: int
     mtime_ns: int
     hexdigest: str = ""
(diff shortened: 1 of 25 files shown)


kmichel-aiven:
There's a typo, I will undraft once fixed, but this is still reviewable.

joelynch (approved):
These are okay as temporary measures.

PR #179, opened 2024-01-31
cassandra: drop nodes from cluster when they are stopped

sjamgade:
cassandra is going to try and contact these stopped nodes to get hashes
from them, but that is bound to fail as the cassandra there has been
stopped. This only delays the whole restore step and causes un-necessary
timouts and various places.

### DIFF
--- a/astacus/coordinator/plugins/cassandra/restore_steps.py
+++ b/astacus/coordinator/plugins/cassandra/restore_steps.py
@@ -200,11 +200,26 @@ class StartCassandraStep(Step[None]):
 @dataclass
 class WaitCassandraUpStep(Step[None]):
     duration: int
+    # Only StopReplacedNodesStep's result can be directly substracted from cluster.nodes
+    # thus retricting the type to Optional[StopReplacedNodesStep]
+    replaced_node_step: Optional[Type[StopReplacedNodesStep]] = None
 
     async def run_step(self, cluster: Cluster, context: StepsContext) -> None:
         last_err = None
+
+        nodes = None
+        if self.replaced_node_step:
+            replaced_nodes = context.get_result(self.replaced_node_step)
+            if len(replaced_nodes) != 1:
+                raise StepFailedError(f"Multiple node replacement not supported: {replaced_nodes}")
+            replaced_node = replaced_nodes[0]
+            if replaced_node not in cluster.nodes:
+                logger.warning("Replaced %s is not part of the cluster.nodes: %s", replaced_node, cluster.nodes)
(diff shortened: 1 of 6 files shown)


sjamgade:
I wonder if the `StopReplacedNodesStep` can return `[]` for example when new node is being added to the cluster when the node count changes (for example). 
ofcouse the example wont be a "restore" but could there be any other scenario

astacus/coordinator/plugins/cassandra/restore_steps.py
```diff
@@ -203,8 +203,14 @@ class WaitCassandraUpStep(Step[None]):
 
     async def run_step(self, cluster: Cluster, context: StepsContext) -> None:
         last_err = None
+
+        replaced_nodes = context.get_result(StopReplacedNodesStep)
```
dmitry-potepalov, line 207: there's no StopReplacedNodes step on full restore, this would just fail with an exception in that case. Maybe instead of excluding the stopped nodes include only the new nodes? in full restore that'd be the whole cluster, in partial restore - only the nodes we restored.

astacus/coordinator/plugins/cassandra/utils.py
```diff
@@ -41,10 +41,14 @@ async def run_subop(
     return await cluster.wait_successful_results(start_results=start_results, result_class=result_class)
 
 
-async def get_schema_hash(cluster: Cluster) -> Tuple[str, str]:
+async def get_schema_hash(cluster: Cluster, exclude_nodes: Optional[list[CoordinatorNode]] = None) -> Tuple[str, str]:
```
dmitry-potepalov, line 44: the pattern in other places is to specify an optional list of nodes we want this to run on, not to exclude a subset. Maybe adding a `nodes: Optional[list[CoordinatorNode]]` and pass that to run_subop?

dmitry-potepalov (changes requested):
Doesn't work on full cluster restore, some other approach is necessary.

astacus/coordinator/plugins/cassandra/restore_steps.py
sjamgade: in full cluster restore there wont be any stopped nodes, so pinging all cluster.nodes should be okay (?)
so took an alternate approach of looking at the Stop step or just follow the defaults

dmitry-potepalov (approved):
LGTM, although I would have done it differently (by including only partial restore nodes, not excluding stopped ones) :) but let's see, I don't think I have a compelling enough argument for my solution over this one.

PR #181, opened 2024-02-11
Upgrade type annotations and dependencies

kmichel-aiven:
Minor cleanup to keep up with more recent python style and fix related linter warnings.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -42,7 +42,7 @@ repos:
           - --rcfile=.pylintrc
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.0.0
+    rev: v1.8.0
     hooks:
       - id: mypy
         exclude: ^setup.py|^vendor/|^astacus/proto/
@@ -50,6 +50,7 @@ repos:
           - types-PyYAML>=6.0.12.2
           - types-requests>=2.28.11.5
           - types-tabulate>=0.9.0.0
+          - types-ujson>=5.9.0.0
           - types-urllib3>=1.26.25.4
           - typing_extensions>=3.10.0
-          - pydantic==1.10.2
+          - pydantic==1.10.9
--- a/astacus/common/cassandra/config.py
+++ b/astacus/common/cassandra/config.py
@@ -17,9 +17,10 @@ https://github.com/samuelcolvin/pydantic/issues/3376
 """
 
 from astacus.common.utils import AstacusModel
+from collections.abc import Sequence
 from pathlib import Path
 from pydantic import root_validator
-from typing import List, Optional
+from typing import Optional
 
 import yaml
 
@@ -29,19 +30,19 @@ BACKUP_GLOB = "data/*/*/backups/"
 
 
 class CassandraClientConfiguration(AstacusModel):
-    config_path: Optional[Path]
+    config_path: Optional[Path] = None
 
     # WhiteListRoundRobinPolicy contact points
-    hostnames: Optional[List[str]]
+    hostnames: Optional[Sequence[str]] = None
 
-    port: Optional[int]
+    port: Optional[int] = None
 
     # PlainTextAuthProvider
     username: str
     password: str
 
     # If set, configure ssl access configuration which requires the ca cert
-    ca_cert_path: Optional[str]
+    ca_cert_path: Optional[str] = None
 
     @classmethod
     @root_validator
@@ -54,7 +55,7 @@ class CassandraClientConfiguration(AstacusModel):
             return self.port
         return int(self.get_config()["native_transport_port"])
 
-    def get_hostnames(self) -> List[str]:
+    def get_hostnames(self) -> Sequence[str]:
         if self.hostnames:
             return self.hostnames
         return ["127.0.0.1"]
--- a/astacus/common/cassandra/config.py
+++ b/astacus/common/cassandra/config.py
@@ -20,7 +20,6 @@ from astacus.common.utils import AstacusModel
 from collections.abc import Sequence
 from pathlib import Path
 from pydantic import root_validator
-from typing import Optional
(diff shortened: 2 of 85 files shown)


astacus/common/cassandra/config.py
```diff
@@ -29,19 +29,19 @@
 
 
 class CassandraClientConfiguration(AstacusModel):
-    config_path: Optional[Path]
+    config_path: Path | None = None
 
     # WhiteListRoundRobinPolicy contact points
-    hostnames: Optional[List[str]]
```
Khatskevich, line 35: this requires python larger then 3.10. Probably it is not stated in required packages

Khatskevich (changes requested):
Just one comment. The rest looks good

astacus/common/cassandra/config.py
kmichel-aiven: It's already specified 3.11 in all places: https://github.com/Aiven-Open/astacus/commit/55c97cb0ba41159c004327974110943187ef5ec4

.pre-commit-config.yaml
```diff
@@ -42,14 +42,15 @@ repos:
           - --rcfile=.pylintrc
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.0.0
+    rev: v1.8.0
     hooks:
       - id: mypy
         exclude: ^setup.py|^vendor/|^astacus/proto/
         additional_dependencies:
```
joelynch, line 49: Just in general as well, why are these not just part of requirements.texting.txt?  Some seem to be defined in multiple places.

mypy.ini
```diff
@@ -1,6 +1,5 @@
 [mypy]
 python_version = 3.11
-plugins = pydantic.mypy
```
joelynch, line 3: Why are we removing this?

joelynch (changes requested):
Mostly nits, but I also have some questions about why the mypy pydantic plugin is removed and also if we should just actually add ujson as a dependency, and if there is a way to consolidate our dev dependencies so that they are only specified in one place.

astacus/node/config.py
```diff
@@ -62,15 +61,15 @@ class NodeConfig(AstacusModel):
     # Where do we hardlink things from the file hierarchy we care about
     # By default, .astacus subdirectory is created in root if this is not set
     # Directory is created if it does not exist
-    root_link: Optional[Path]
+    root_link: Path | None = None
```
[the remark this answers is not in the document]
joelynch: I guess it has to be added if we remove use of the pydantic plugin.  There is other weirdness from removing the plugin though I think.

mypy.ini
kmichel-aiven: Because the plugin was crashing, I forgot to actually find a solution.

Bumping to pydantic 1.10.9 Instead of 1.10.2 fixed it.

.pre-commit-config.yaml
kmichel-aiven: Because pre-commit runs in its own venv managed by its yaml file.

It's a wontfix: https://github.com/pre-commit/pre-commit/issues/730

astacus/config.py
```diff
@@ -49,22 +48,22 @@ class GlobalConfig(AstacusModel):
...
-def get_config_content_and_hash(config_path: Union[str, Path]) -> Tuple[str, str]:
+def get_config_content_and_hash(config_path: str | Path) -> tuple[str, str]:
     with open(config_path, "rb") as fh:
         config_content = fh.read()
     config_hash = hashlib.sha256(config_content).hexdigest()
     return config_content.decode(), config_hash
 
 
-def set_global_config_from_path(app: FastAPI, path: Union[str, Path]) -> GlobalConfig:
+def set_global_config_from_path(app: FastAPI, path: str | Path) -> GlobalConfig:
```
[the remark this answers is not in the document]
kmichel-aiven: [`os.PathLike`](https://docs.python.org/3/library/os.html#os.PathLike) already exists and is something very different, I'd rather not introduce a homonym type (and I don't really think `str` is anything like a `Path`.)

astacus/coordinator/plugins/flink/steps.py
```diff
@@ -37,8 +37,10 @@ async def run_step(self, cluster: Cluster, context: StepsContext) -> Dict[str, A
                 raise TransientException("Concurrent modification during access entities retrieval")
         return results
 
-    async def _get_result(self, zk_path: str, connection: ZooKeeperConnection, change_watch: ChangeWatch) -> Dict[str, Any]:
-        result: Dict[str, Any] = {}
+    async def _get_result(
+        self, zk_path: str, connection: ZooKeeperConnection, change_watch: ChangeWatch
+    ) -> Mapping[str, Any]:
```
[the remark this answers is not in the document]
kmichel-aiven: For this one, dict is kind of reasonable, the important part was that the return value of step should really be Mutable (we don't want mutable values insides the Context).

astacus/server.py
```diff
@@ -29,7 +28,7 @@
 
 logger = logging.getLogger(__name__)
 
-app: Optional[FastAPI] = None
+app: FastAPI | None = None
```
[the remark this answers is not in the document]
kmichel-aiven: Sounds good, but in a separate PR.

astacus/common/op.py
```diff
@@ -41,7 +41,7 @@ class Status(Enum):
     class Info(AstacusModel):
         op_id: int = 0
         op_name: str = ""
-        op_status: Optional["Op.Status"]
+        op_status: Optional["Op.Status"] = None
```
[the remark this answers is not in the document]
kmichel-aiven: I want to simplify this and lift the inner classes outside of class `Op` (which lets us remove the quotes) but I didn't want to expand this PR too much.

PR #182, opened 2024-02-19
Storage refactor in prep for per node manifests

joelynch:
tldr: use composition over inheritance.

## Context

Currently `Storage` uses inheritance and mixins heavily.  Methods for accessing backup manifests ("jsons") as well as actual file contents ("hexdigests") are all together in a single large class.

A change we have talked about is per node manifests, ie a list of all the files on each node for each manifest.  Further adding a new mixin for the node manifests to `Storage` would be unwieldy and be too much to have in one class.  So as preparation, I am splitting the `Storage` class.

## Changes

Split Storage into smaller, composable classes.
* `Storage` and it's implementations are the "storage backend".  They handle actually reading and writing from object storage / the file system / memory / anywhere else.  Caching is also transparently handled by this backend class, see `CachedStorage`.
* `MultiStorage` is similar to before.  It routes access to multiple `Storage` classes keyed by name.
* `JsonStore` and `HexDigestStore` now wrap a `Storage` class rather than mixing in.  Since they are not a `Storage` class themselves, I changed the suffix (if someone has a better name pls let me know).
* `StorageManager` combines all these things to provide access to `JsonStore` and `HexDigestStore` for both the node and the coordinator APIs.

Some other small changes:
* The current inheritance model for `ThreadLocalStorage` won't work well if we need to cache multiple things instead of just one class, which will be necessary for `Uploader` with per node manifests.  Use composition here, renaming to `CopiedThreadLocal`.
* Use `tmp_path: Path` over `tmpdir: py.path.local`


joelynch:
aiven-core tests ran successfully here https://github.com/aiven/aiven-core/pull/44634

PR #183, opened 2024-02-27
replace pydantic with msgspec for ipc [DDB-881]

kmichel-aiven:
This isn't ideal but still greatly improves memory usage and should enable us
to deal with larger backups while working on a better solution for even larger backups.

The main goal is to replace `BackupManifest` and its components, but that still requires
migrating all the models for the API request/response (or creating duplicates
and copying between the two, which would waste all the saved memory usage).

The config was spared and kept as pydantic, to avoid dealing with rohmu models,
this required splitting the Retention model that was both used for config and ipc.

While not perfect, this still reduces the peak memory usage  from 526MiB to 217MiB
during backups and from 563MiB to 210MiB during restore (using a test procedure
with a ClickHouse backups containing many partitions).

A part of that memory is constant usage from python itself, so the savings for larger
backups are expected to be more than 60%.

### DIFF
--- a/astacus/common/rohmustorage.py
+++ b/astacus/common/rohmustorage.py
@@ -6,10 +6,10 @@ See LICENSE for details
 Rohmu-specific actual object storage implementation
 
 """
-from .storage import Json, MultiStorage, Storage, StorageUploadResult
+from .storage import MultiStorage, Storage, StorageUploadResult
 from .utils import AstacusModel, fifo_cache
 from astacus.common import exceptions
-from collections.abc import Mapping
+from collections.abc import Iterator, Mapping
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
 from cryptography.hazmat.primitives.asymmetric.rsa import RSAPrivateKey, RSAPublicKey
@@ -21,9 +21,10 @@ from rohmu.encryptor import EncryptorStream
 from rohmu.typing import Metadata
 from typing import BinaryIO, TypeAlias
 
+import contextlib
 import io
-import json
 import logging
+import mmap
 import os
 import rohmu
 import tempfile
@@ -199,23 +200,26 @@ class RohmuStorage(Storage):
         key = os.path.join(self.json_key, name)
         self.storage.delete_key(key)
 
-    def download_json(self, name: str) -> Json:
+    @contextlib.contextmanager
+    def open_json_bytes(self, name: str) -> Iterator[mmap.mmap]:
         key = os.path.join(self.json_key, name)
-        f = io.BytesIO()
-        self._download_key_to_file(key, f)
-        f.seek(0)
-        return json.load(f)
+        with tempfile.TemporaryFile(dir=self.config.temporary_directory) as temp_file:
+            self._download_key_to_file(key, temp_file)
+            temp_file.seek(0)
+            with mmap.mmap(temp_file.fileno(), 0, access=mmap.ACCESS_READ) as mapped_file:
+                yield mapped_file
 
     def list_jsons(self) -> list[str]:
         return self._list_key(self.json_key)
 
-    def upload_json_str(self, name: str, data: str) -> bool:
+    def upload_json_bytes(self, name: str, data: bytes | mmap.mmap) -> bool:
         key = os.path.join(self.json_key, name)
-        f = io.BytesIO()
-        encoded_data = data.encode()
-        f.write(encoded_data)
-        f.seek(0)
-        self._upload_key_from_file(key, f, len(encoded_data))
+        if isinstance(data, bytes):
+            f = io.BytesIO(data)
+            self._upload_key_from_file(key, f, len(data))
+        else:
+            data.seek(0)
+            self._upload_key_from_file(key, data, len(data))
         return True
 
 
--- a/astacus/common/utils.py
+++ b/astacus/common/utils.py
@@ -12,13 +12,15 @@ from __future__ import annotations
 
 from abc import ABC
 from collections import deque
+from collections.abc import AsyncIterable, AsyncIterator, Callable, Hashable, Iterable, Mapping
 from contextlib import contextmanager
 from multiprocessing.dummy import Pool  # fastapi + fork = bad idea
 from pathlib import Path
 from pydantic import BaseModel
-from typing import Any, AsyncIterable, Callable, Final, Generic, Hashable, Iterable, Mapping, TypeAlias, TypeVar
+from typing import Any, Final, Generic, TypeAlias, TypeVar
 
 import asyncio
+import contextlib
 import datetime
 import httpcore
 import httpx
@@ -133,6 +135,41 @@ async def httpx_request(
         except httpx.HTTPError as ex:
             logger.warning("Unexpected response from %s to %s: %r", url, caller, ex)
         return None
+
+
+@contextlib.asynccontextmanager
+async def httpx_request_stream(
+    url: str | httpx.URL,
+    *,
+    caller: str,
+    method: str = "get",
+    timeout: float = 10.0,
+    ignore_status_code: bool = False,
+    **kw,
+) -> AsyncIterator[httpx.Response | None]:
+    """Wrapper for httpx.request which handles timeouts as non-exceptions,
+    and returns only valid results that we actually care about.
+    """
+    # TBD: may need to redact url in future, if we actually wind up
+    # using passwords in urls here.
+    logger.info("async-request %s %s by %s", method, url, caller)
+    async with httpx.AsyncClient() as client:
+        try:
+            async with client.stream(method, url, timeout=timeout, **kw) as r:
+                if not r.is_error or ignore_status_code:
+                    yield r
+                    return
+                await r.aread()
+                logger.warning("Unexpected response status code from %s to %s: %s %r", url, caller, r.status_code, r.text)
+        except (httpcore.NetworkError, httpcore.TimeoutException) as ex:
+            # Unfortunately at least current httpx leaks these
+            # exceptions without wrapping them. Future versions may
+            # address this hopefully. I believe httpx.TransportError
+            # replaces it in future versions once we upgrade.
+            logger.warning("Network error from %s to %s: %r", url, caller, ex)
+        except httpx.HTTPError as ex:
+            logger.warning("Unexpected response from %s to %s: %r", url, caller, ex)
+        yield None
 
 
 def build_netloc(host: str, port: int | None = None) -> str:
--- a/astacus/coordinator/cleanup.py
+++ b/astacus/coordinator/cleanup.py
@@ -22,10 +22,21 @@ class CleanupOp(SteppedCoordinatorOp):
 
     def __init__(self, *, c: Coordinator, req: ipc.CleanupRequest) -> None:
         context = c.get_operation_context()
-        retention = c.config.retention.copy()
-        if req.retention is not None:
-            # This returns only non-defaults -> non-Nones
-            for k, v in req.retention.dict().items():
-                setattr(retention, k, v)
+        if req.retention is None:
+            retention = ipc.Retention(
+                minimum_backups=c.config.retention.minimum_backups,
+                maximum_backups=c.config.retention.maximum_backups,
+                keep_days=c.config.retention.keep_days,
+            )
+        else:
+            retention = ipc.Retention(
+                minimum_backups=coalesce(req.retention.minimum_backups, c.config.retention.minimum_backups),
+                maximum_backups=coalesce(req.retention.maximum_backups, c.config.retention.maximum_backups),
+                keep_days=coalesce(req.retention.keep_days, c.config.retention.keep_days),
+            )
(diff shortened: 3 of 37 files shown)


astacus/common/utils.py
```diff
@@ -135,6 +136,41 @@ async def httpx_request(
...
+    # TBD: may need to redact url in future, if we actually wind up
+    # using passwords in urls here.
+    logger.info("async-request %s %s by %s", method, url, caller)
+    async with httpx.AsyncClient() as client:
+        try:
+            async with client.stream(method, url, timeout=timeout, **kw) as r:
+                if not r.is_error or ignore_status_code:
+                    yield r
+                    return
+                await r.aread()
```
joelynch, line 161: This is a weird api, so aread sets the `text` field

kmichel-aiven: Yes, this looks weird.
That's probably because text is not async (async properties are not a thing), and they don't want to block unexpectedly while reading `text`, instead it raises and exception if `aread` was not already called.

astacus/coordinator/cluster.py
```diff
@@ -198,16 +199,20 @@ async def wait_successful_results(
...
+                        failures[i] += 1
+                        if failures[i] >= self.poll_config.maximum_failures:
+                            raise WaitResultError("too many failures")
+                        continue
+                    # We got something -> decode the result
+                    assert isinstance(r, httpx.Response)
+                    payload = bytearray()
+                    async for chunk in r.aiter_bytes():
+                        payload.extend(chunk)
+                result = msgspec.json.decode(payload, type=result_class)
```
joelynch, line 215: I'm not seeing any performance advantage from using stream here.  We are just collecting into a `bytearray` anyway - might as well let the library do this.

astacus/coordinator/coordinator.py
```diff
@@ -111,15 +111,17 @@ def delete_json(self, name: str) -> None:
...
+    @contextlib.contextmanager
+    def open_json_bytes(self, name: str) -> Iterator[bytearray]:
+        with self.storage.open_json_bytes(name) as json_bytes:
+            yield json_bytes
 
     def list_jsons(self) -> list[str]:
         return self.storage.list_jsons()
 
-    def upload_json_str(self, name: str, data: str) -> bool:
+    def upload_json_bytes(self, name: str, data: bytes | bytearray) -> bool:
```
joelynch, line 122: note `bytearray` is a subclass of `bytes`.


```suggestion
    def upload_json_bytes(self, name: str, data: bytes) -> bool:
```

astacus/common/rohmustorage.py
```diff
@@ -185,23 +186,24 @@ def delete_json(self, name: str) -> None:
...
+            with mmap.mmap(temp_file.fileno(), 0, access=mmap.ACCESS_READ) as mapped_file:
+                # https://docs.python.org/3/library/mmap.html
+                # > Memory-mapped file objects behave like both bytearray and like file objects.
+                yield mapped_file  # type: ignore
 
     def list_jsons(self) -> list[str]:
         return self._list_key(self.json_key)
 
-    def upload_json_str(self, name: str, data: str) -> bool:
+    def upload_json_bytes(self, name: str, data: bytes | bytearray) -> bool:
```
joelynch, line 203:
```suggestion
    def upload_json_bytes(self, name: str, data: bytes) -> bool:
```

astacus/coordinator/list.py
```diff
@@ -4,23 +4,14 @@
 
 """
 from astacus.common import ipc, magic
-from astacus.common.json_view import (
```
joelynch, line 7: Lets remove `json_view` since we no longer work with json directly.

Reduce the amount of code we need to maintain.

astacus/coordinator/manifest.py
```diff
@@ -5,10 +5,26 @@
...
 
+import msgspec
+
 
 async def download_backup_manifest(json_storage: asyncstorage.AsyncJsonStorage, backup_name: str) -> ipc.BackupManifest:
-    d = await json_storage.download_json(backup_name)
-    manifest = await run_in_threadpool(ipc.BackupManifest.parse_obj, d)
+    def download_manifest() -> ipc.BackupManifest:
+        with json_storage.storage.open_json_bytes(backup_name) as manifest_content:
+            return msgspec.json.decode(manifest_content, type=ipc.BackupManifest)
```
joelynch, line 14: Might be time to retire `AsyncJsonStorage` IMO if we are bypassing the abstraction like this.

astacus/node/download.py
```diff
@@ -195,3 +192,19 @@ def download(self) -> None:
...
+
+
+class Skip(msgspec.Struct):
+    pass
+
+
+def download_snapshot(json_storage: JsonStorage, backup_name: str, index: int) -> ipc.SnapshotResult:
+    snapshot_results = msgspec.defstruct(
+        "SnapshotResults",
+        [*[(f"skip_{i}", Skip) for i in range(index)], ("value", ipc.SnapshotResult)],
```
joelynch, line 204: Fancy

astacus/common/rohmustorage.py
```diff
@@ -185,23 +186,24 @@ def delete_json(self, name: str) -> None:
...
+    def open_json_bytes(self, name: str) -> Iterator[bytearray]:
         key = os.path.join(self.json_key, name)
-        f = io.BytesIO()
-        self._download_key_to_file(key, f)
-        f.seek(0)
-        return json.load(f)
+        with tempfile.TemporaryFile(dir=self.config.temporary_directory) as temp_file:
+            self._download_key_to_file(key, temp_file)
+            temp_file.seek(0)
+            with mmap.mmap(temp_file.fileno(), 0, access=mmap.ACCESS_READ) as mapped_file:
```
joelynch, line 195: Considering we have ACCESS_READ here only, having the return type be `bytes` might make more sense actually.

tests/unit/common/test_storage.py
```diff
@@ -101,11 +104,12 @@ def test_caching_storage(tmpdir: py.path.local, mocker: MockerFixture) -> None:
...
 
     # We shouldn't wind up in download method of rohmu at all.
-    mockdown = mocker.patch.object(storage.backend_storage, "download_json")
+    mockdown = mocker.patch.object(storage.backend_storage, "open_json_bytes")
     # Nor list hexdigests
     mocklist = mocker.patch.object(storage.backend_storage, "list_jsons")
 
-    assert storage.download_json(TEST_JSON) == TEST_JSON_DATA
+    with storage.open_json_bytes(TEST_JSON) as json_bytes:
+        assert json.loads(bytes(json_bytes)) == TEST_JSON_DATA
```
joelynch, line 112: This is a bit sad that we are faking types like this that can have confusing errors.

joelynch (dismissed):

astacus/coordinator/cleanup.py
```diff
@@ -22,10 +22,21 @@ async def create(*, c: Coordinator = Depends(), req: ipc.CleanupRequest = ipc.Cl
...
     def __init__(self, *, c: Coordinator, req: ipc.CleanupRequest) -> None:
         context = c.get_operation_context()
-        retention = c.config.retention.copy()
-        if req.retention is not None:
-            # This returns only non-defaults -> non-Nones
-            for k, v in req.retention.dict().items():
-                setattr(retention, k, v)
+        if req.retention is None:
+            retention = ipc.Retention(
+                minimum_backups=c.config.retention.minimum_backups,
```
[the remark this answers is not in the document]
kmichel-aiven: The config is still using pydantic, to avoid reaching up to the rohmu pydantic models that are external to astacus).
It's unpleasant to have two Retention types, but less than duplicating all the rohmu models (and keeping them in sync with the external definition).

astacus/coordinator/cluster.py
kmichel-aiven: Using the stream API reduces the memory usage: without that, the http library does extra copies when merging chunks, and it keeps a separate copy of the entire content attached to the response object.

astacus/coordinator/coordinator.py
kmichel-aiven:
```>>> issubclass(bytearray, bytes)
False
```
?

astacus/coordinator/list.py
kmichel-aiven: Yes, I want to do that in a separate PR to avoid huge line change count.

astacus/coordinator/manifest.py
kmichel-aiven: Yes, but in a separate PR.

astacus/common/rohmustorage.py
kmichel-aiven: I think you meant that in a context where bytearray is a subclass of bytes, but since it's not the case, I don't think we can.

tests/unit/common/test_storage.py
kmichel-aiven: Where are we faking types ? (the call to bytes is not a C cast, it's a real conversion to a bytes object)

astacus/coordinator/coordinator.py
joelynch: I mean in terms of mypy.  subclass is the wrong word but the type `bytearray` is accepted everywhere `bytes` is and you can write stuff like this

```python
test: bytes = bytearray()
```

tests/unit/common/test_storage.py
joelynch: mypy thinks `json_bytes` here is a bytearray, which `json.loads` accepts.  But in fact it is a mmap object and `json.loads` fails with a `TypeError`.

astacus/coordinator/coordinator.py
joelynch: Subtype is the word.

tests/unit/common/test_storage.py
kmichel-aiven: Aah, you're right that's a bad typing issue. I didn't realize I did that, I'll adjust the types :+1: 

(That might mean I'll have to expose the mmap type everywhere, but I'd rather have truthful code even if the truth isn't pretty.)

astacus/coordinator/cluster.py
joelynch: > the http library does extra copies when merging chunks

When using the non stream API it is just doing `self._content = b"".join([part async for part in self.stream])` which is basically just what you've implemented, except maybe more efficient in terms of copying (if `len(part)` is large enough).

https://github.com/encode/httpx/blob/master/httpx/_models.py#L422

There might be a very small memory benefit to your code over theirs since it doesn't have to have an extra array of references to `bytes`.

> and it keeps a separate copy of the entire content attached to the response object

Not a copy just a reference.

tests/unit/common/test_storage.py
joelynch: Yea agreed I would go with the truth even if it is ugly

astacus/coordinator/cluster.py
kmichel-aiven: I got a little confused, because you're right that `_content` is just a reference, but I do see extra memory usage in the profiler without that change.

(This change is actually in a separate commit, because first I tried without, then I saw the `b"".join[...]` still allocated during peak memory usage, and that disappeared once I switched to the stream approach).

I assumed it was because the reference to the response was somehow kept alive for longer than expected (that wouldn't be completely unexpected, especially when you have things like sentry that hook themselves everywhere to collect info about requests). But I was wrong, that's not what happens at all.

After checking the size and number of chunks we have for large response and after digging through cpython, the extend is really more efficient, but what really happens is:

In the `bytearray.extend` case:
 - httpx allocates and fetches a 64k chunk of response in a `bytes`, yields it
 - `bytearray_extend` is passed that chunk, extends its internal buffer with `realloc` (with a multiplicative factor of 1.125, so progressively appending 1000 chunks of 64k does not result in 1000 reallocs, it only does 60 reallocs, and the system `realloc` for large sizes is itself smart about avoiding copies anyway).
 - the 64k chunk is now useless, it has zero refs and can be freed (that memory will be reused for the next block).
 - repeat that for each chunk until the response is complete.

In the `b"".join[...]` case:
 - to allocate the output buffer, `stringlib_bytes_join` needs to know the size of all chunks that it needs to join together.
 - so it has to consume the entire iterable of all chunks.
 - this allocates all N chunks of 64k bytes.
 - then it can allocate a large buffer of N*64k bytes.
 - then it copies everything from all the small chunks to the large buffer.
 - finally it can free all chunks.

So, effectively, the join will use ~twice as much memory, because at some point it will retain both the joined response and all the small chunks at the same time, while extend will only retain one chunk at a time.

astacus/coordinator/coordinator.py
kmichel-aiven: This doesn't apply anymore with things typed with mmap.

astacus/common/rohmustorage.py
kmichel-aiven: This doesn't really apply anymore with things typed as mmap.

kmichel-aiven:
Fixing the conflict now, should be quick.

aris-aiven (approved):
LGTM, didn't find anything that would cause a regression from the current state

PR #186, opened 2024-03-04
reuse individual backup list entries

kmichel-aiven:
A common sequence of operations is to first list backups, then restore one of the listed backups, and then monitor the list of available backups with regular polling.

Since both the restore operation and the uncached list operation can load manifests, we'd rather not have the combined memory usage of both operations at the same time.

If the restore operation takes longer than the list response TTL (one hour with the default configuration), then a regular polling of the backups list will cause that simultaneous memory usage right after the cached response expires.

There is value in expiring the entire response after some time, because the list of backups can change without astacus being notified of that change.

However, the details about a single backup are not supposed to change, backups are immutable (until they are deleted).

Even when the list response is expired by TTL, reuse individual entries from the old response (if the entry still exists in the new response).

With that modification, the cached list response will expire, but since the individual entries will not be expired, the new list response will be built by listing the backup names from object storage but without loading the manifests.

Ideally, we should pre-compute all the memory-intensive values when creating the backup instead of doing it again every time we list the backups. But we want to improve the restore operation for backups that already exist.

This change also improves the "normal" flow of operation even when restore are not involved: instead of re-loading all manifests every hour, we will only load the most recently created backup once, right after the backup is created.

[DDB-881]


tests/unit/coordinator/test_list.py
```diff
@@ -228,7 +229,7 @@ def test_api_list_deduplication(backup_manifest: BackupManifest, tmpdir: PathLik
     storage.upload_hexdigest_bytes("FAKEDIGEST", b"fake-digest-data")
 
     list_request = ListRequest(storage="x")
-    list_response = list_backups(req=list_request, json_mstorage=multi_rohmu_storage)
+    list_response = list_backups(req=list_request, json_mstorage=multi_rohmu_storage, cache={})
```
Khatskevich, line 232: should there be some test added to make sure cache is used properly?

astacus/coordinator/api.py
```diff
@@ -142,6 +143,13 @@ async def _list_backups(
     return StructResponse(list_response)
 
 
+def get_cached_entries(list_response: ipc.ListResponse) -> CachedListEntries:
```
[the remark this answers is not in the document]
kmichel-aiven: I've made the function name more specific.

tests/unit/coordinator/test_list.py
kmichel-aiven: Good point, I've added some positive and negative tests.

kmichel-aiven:
I had to add an extra commit to fix CI, ZooKeeper 3.8.3 was removed from the CDN.

Khatskevich (approved):

PR #188, opened 2024-03-11
clickhouse: make restore work with annoy index

kmichel-aiven:
Like other experimental settings, whether they should be allowed or not is decided somewhere else, but astacus should not unnecessary fail if they were allowed.

[DDB-500]


alanfranz (approved):
LGTM

PR #191, opened 2024-03-11
tests: use asyncio_mode = auto [DDB-901]

joelynch:
This means that we do not miss tests by accidentally forgetting the pytest mark.  Additionally this is needed to run tests using aiven-core pytest.

Additionally, I removed a duplicate github actions workflow.


alanfranz (dismissed):
```
tests/unit/coordinator/plugins/test_base.py::test_upload_step_uses_new_request_if_supported[no_feature] SKIPPED (async def function and no async plugin installed (see warnings))
tests/unit/coordinator/plugins/test_base.py::test_upload_step_uses_new_request_if_supported[validate_file_hashes] SKIPPED (async def function and no async plugin installed (see warnings))
```

So, async tests don't run nevertheless; this _could_ be the intended behaviour (maybe we just want to run those in aiven-core) but the commit message is misleading, as it would imply that we'd be running those tests by now; so I think we should either add an async plugin or amend the commit and pr message.

PR #192, opened 2024-03-15
Properly support materialized views restoration

Khatskevich:
This pr enables the restoration of databases with materialized views when:
1. Source table was deleted before backup.
2. Source table was not backed up by Astacus.

This is achieved by using `ATTACH` instead of `CREATE`, which performs
fewer checks at the moment of creation.

Misc: add a test for basic views to make sure they are restored properly
in similar case.

[DDB-890]

### DIFF
--- a/tests/integration/coordinator/plugins/clickhouse/test_plugin.py
+++ b/tests/integration/coordinator/plugins/clickhouse/test_plugin.py
@@ -9,6 +9,7 @@ from astacus.coordinator.plugins.base import OperationContext
 from astacus.coordinator.plugins.clickhouse.client import (
     ClickHouseClient,
     ClickHouseClientQueryError,
+    escape_sql_identifier,
     escape_sql_string,
     HttpClickHouseClient,
 )
@@ -143,6 +144,7 @@ async def restored_cluster_manager(
                             "restorable",
                         )
                     run_astacus_command(astacus_cluster, "restore", "--storage", "restorable")
+                    await sync_replicated_tables(clients)
                     yield clients
 
 
@@ -173,6 +175,18 @@ async def fixture_function_restored_cluster(
             restorable_cluster, ports, None, clickhouse_restore_command, function_minio_bucket
         ) as clients:
             yield clients
+
+
+async def sync_replicated_tables(clients: Sequence[ClickHouseClient]) -> None:
+    # Get replicated tables to sync
+    rows = await clients[0].execute(
+        b"SELECT name FROM system.tables WHERE database = 'default' AND engine like 'Replicated%'"
+    )
+    for client in clients:
+        for row in rows:
+            table_name = row[0]
+            assert isinstance(table_name, str)
+            await client.execute(f"SYSTEM SYNC REPLICA default.{escape_sql_identifier(table_name.encode())} STRICT".encode())
 
 
 async def setup_cluster_content(clients: Sequence[HttpClickHouseClient], use_named_collections: bool) -> None:
--- a/tests/integration/coordinator/plugins/clickhouse/test_plugin.py
+++ b/tests/integration/coordinator/plugins/clickhouse/test_plugin.py
@@ -6,9 +6,10 @@ See LICENSE for details
 from _pytest.fixtures import SubRequest
 from astacus.common.ipc import RestoreRequest
 from astacus.coordinator.plugins.base import OperationContext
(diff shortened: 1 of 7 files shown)


tests/integration/coordinator/plugins/clickhouse/test_plugin.py
```diff
@@ -188,7 +188,6 @@ async def setup_cluster_content(clients: Sequence[HttpClickHouseClient], use_nam
     await clients[0].execute(
         b"CREATE TABLE default.mysql (a Int) ENGINE = MySQL('https://host:1234', 'database', 'table', 'user', 'password')"
     )
-    await clients[0].execute(b"CREATE TABLE default.s3 (a Int) ENGINE = S3('http://bucket.s3.amazonaws.com/key.json')")
```
kmichel-aiven, line 191: Can you make that dependent on the presence of the engine instead of disabling it ?

(By using `SELECT name FROM system.table_engines`)

kmichel-aiven (changes requested):

tests/integration/coordinator/plugins/clickhouse/test_plugin.py
Khatskevich: I have fixed this.
But I have added incomplete set of table engines and did not used it in `Table`. Is it fine? (not sure if enum should not be used in `Table` because some new engines can be easily added to Clickhouse)

tests/integration/coordinator/plugins/clickhouse/test_plugin.py
```diff
@@ -343,6 +377,50 @@ async def test_restores_materialized_view_data(restored_cluster: Sequence[ClickH
...
+            await client.execute(b"SELECT thekey2 FROM default.view_deleted_source ORDER BY thekey2")
+        assert raised.value.status_code == 404
+        assert raised.value.exception_code == unknown_table_exception_code
+    # Recreated deleted table works again
+    await restored_cluster[0].execute(
+        b"CREATE TABLE default.source_table_for_view_deleted (thekey UInt32, thedata String) "
+        b"ENGINE = ReplicatedMergeTree ORDER BY (thekey)"
+    )
+    await restored_cluster[0].execute(b"INSERT INTO default.source_table_for_view_deleted VALUES (11, '11')")
+    await restored_cluster[2].execute(b"INSERT INTO default.source_table_for_view_deleted VALUES (17, '17')")
```
kmichel-aiven, line 415: Sorry I didn't notice earlier.

This test will mutate the `restored_cluster` that has a module scope, this will affect other tests.

This needs a separate `function_restored_cluster` fixture which does the same as the module fixture but function-scoped and without the steps parametrization. (We don't want to make all tests function scoped, since the restore is slow and this test doesn't need to check the partial-restore-and-retry that we use for other tests)

kmichel-aiven (changes requested):

tests/integration/coordinator/plugins/clickhouse/test_plugin.py
Khatskevich: Sorry, I did it intentionally, I thought it is fine. But probably it is better to not touch state which looks immutable.
The fix looks a bit ugly, but fine. I am not exactly sure what this test is testing, but let it better be there, as someone will want to `undelete` some table one day.

tests/integration/coordinator/plugins/clickhouse/test_plugin.py
```diff
@@ -343,6 +430,57 @@ async def test_restores_materialized_view_data(restored_cluster: Sequence[ClickH
...
+async def test_restores_view_with_deleted_source_table(restored_cluster: Sequence[ClickHouseClient]) -> None:
+    unknown_table_exception_code = 60
+    for client in restored_cluster:
+        with pytest.raises(ClickHouseClientQueryError) as raised:
+            await client.execute(b"SELECT thekey2 FROM default.view_deleted_source ORDER BY thekey2")
+        assert raised.value.status_code == 404
+        assert raised.value.exception_code == unknown_table_exception_code
+
+
+async def test_restores_view_undeleted_source_table(restored_cluster: Sequence[ClickHouseClient]) -> None:
```
kmichel-aiven, line 468: The other test is using `function_restored_cluster`, but this test, which contains CREATE TABLE, still uses the global `restored_cluster`.

Since it's mutating, it should also use `function_restored_cluster` ?

kmichel-aiven (approved):

PR #193, opened 2024-03-26
Fix astacus/version issues

alanfranz:
Astacus relied on autogenerated versioning from git which could happen or change without real knowledge of the developer, and could cause issues if an accidental git directory was found in a parent when using astacus from a non-git tarball.

This PR:
* fixes accidental version setting
* Tries to untangle automagic version setting in multiple parts of the codebase
* Supports external setting for the desired package version

It seems that we're using all possible packaging files and approaches - multiple requirements files, pyproject, setup.py AND setup.cfg - I didn't want to enter that packaging-related discussion because I couldn't see the end of it, so I just took the existing versioning code and made it predictable & error-proof, I didn't change the "current way of doing things".

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -17,6 +17,12 @@ jobs:
           python-version: '3.11'
           cache: 'pip'
 
+      - name: Install testing requirements
+        run: pip install -r requirements.testing.txt
+
+      - name: generate version
+        run: make generate_version_from_git
+
       - name: Install requirements
         run: pip install -r requirements.txt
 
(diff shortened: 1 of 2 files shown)


aris-aiven (changes requested):
Conflicts need to be fixed since we're now packaging using `pyproject.toml`

alanfranz:
I'm not sure what should I do anymore. It seems that a https://github.com/Aiven-Open/astacus/pull/194 was merged before this and touched the same files. But I don't understand whether it addresses the same concerns.

alanfranz:
it seems this was already solved independently.

PR #194, opened 2024-03-27
Update project to use pyproject.toml

RommelLayco:
This is to ensure that the project meets the standard packing guidelines in PEP-0621, and that the version of the package meets PEP-440


facetoe (approved):

alanfranz:
@RommelLayco automagic version setting for this doesn't seem to work without a .git directory (which is what we're using when build the package in aiven-core). Also, is there any way to manually set the version from "outside" at build time?

And, it seems the gh action was updated but there's no Makefile target to actually create the astacus version.

PR #195, opened 2024-03-28
Fix cluster coordination request usage and corresponding filter

aris-aiven:


astacus/coordinator/cluster.py
```diff
@@ -72,7 +72,7 @@ async def request_from_nodes(
         reqs: Sequence[ipc.NodeRequest] | None = None,
         nodes: Sequence[CoordinatorNode] | None = None,
         **kw,
-    ) -> Sequence[Result | None]:
+    ) -> Sequence[BaseException | httpx.Response | Mapping[str, Any] | None]:
```
kmichel-aiven, line 75: Isn't this equivalent to the `Result` typealias ?

aris-aiven: Correct, was a bit too eager there. Oddly enough, running mypy as part of pre-commit doesn't show the following error (and others), but running it as a local hook does.
```
astacus/coordinator/plugins/base.py:842: error: Incompatible types in assignment (expression has type "Sequence[BaseException | Response | Mapping[str, Any] | None]", variable has type "Sequence[Response | None]")  [assignment]
```
What do you think about replacing the pre-commit mypy hook with a local one for type checking the repo?

kmichel-aiven: I'd rather know why there is a difference, what's the version of the mypy outside of pre-commit ?

(And also, did you have the error when mypy was recently pinned back to 1.0.0. or also with 1.8.0 ?)

aris-aiven: The error is with 1.0.0, 1.8.0 and 1.9.0 when mypy is run in the dev venv (rather than the venv created by pre-commit). So the difference seems to be that pre-commit runs mypy with `--ignore-missing-imports` by default, which hides both missing dependencies in `additional_dependencies` (eg. fastapi, httpx, rohmu) and actual type errors related to these packages...

aris-aiven: Hence why running mypy as a local hook could be useful: we'd only have to specify the dependencies correctly once in `pyproject.toml`. On the down side, we have some errors to fix:
```
Found 25 errors in 10 files (checked 168 source files)
```

kmichel-aiven: Looks like we can both use pre-commit and avoid ignored errors and duplicated dependencies: https://github.com/Aiven-Open/astacus/pull/201

(`make lint` gives me 32 mypy errors with that branch, so it seems to work).

aris-aiven: Great, I'll submit a PR a bit later with the local hook, fixing some errors and ignoring some others for the time being.

aris-aiven:
Closing in favour of https://github.com/Aiven-Open/astacus/pull/210

PR #204, opened 2024-04-02
fix very slow HTTP queries

kmichel-aiven:
httpx has a silly bug where it (very slowly) loads the ssl context even when it's not used at all.

We can work around that by disabling certificates verification when the protocol is HTTP.

A better fix would be to have a proper long-living scope for the client, but this requires large changes in astacus API implementation.


aris-aiven (approved):
Nice find 👍

PR #205, opened 2024-04-02
tests: fix flaky inter-node select after insert

kmichel-aiven:
There are no insert to sync after a restore. On the other hand, we need to sync after the manual inserts in the views and materialized views tests.


aris-aiven (approved):
lgtm

PR #206, opened 2024-04-02
clickhouse: improve error reporting

kmichel-aiven:
Include the full ClickHouse error message in the
ClickHouseClientQueryError expression.

This contains the real cause of any failed query and helps debugging.


aris-aiven (approved):
Great, thank you !

PR #210, opened 2024-04-11
Transition to using a local hook for type checking and fix surfaced type errors

aris-aiven:
The local hook allows specifying the dependencies only once, in the `pyproject.toml` file.
 This ensures they're the same at development and CI times.

Typing file helper functions and Rohmu storage methods requires changing some parameter types to use `IO[bytes]` and `FileLike`. 
These allow pass both `BinaryIO` and `_TemporaryFileWrapper[bytes]` types to the functions.

Some type errors were ignored, notably to work around Pydantic quirks (parsing of enum value, models using a discriminator field to provide union typing).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,19 +11,16 @@ GENERATED_PROTOBUFS = $(patsubst %.proto,%_pb2.py,$(PROTOBUFS))
 
 GENERATED = $(GENERATED_PROTOBUFS)
 
-PYTHON = python3
-DNF_INSTALL = sudo dnf install -y
-
 ZOOKEEPER_VERSION = 3.8.4
 ZOOKEEPER_HASH = 284cb4675adb64794c63d95bf202d265cebddc0cda86ac86fb0ede8049de9187
 
 default: $(GENERATED)
 
 venv: default
-	$(PYTHON) -m venv venv && source venv/bin/activate && pip install -U pip && pip install . ".[dev]"
+	python3 -m venv venv && source venv/bin/activate && pip install -U pip && pip install . ".[dev]"
 
 clean:
-	rm -rf rpm/ $(GENERATED)
+	rm -rf rpm/ $(GENERATED) venv/ .mypy_cache/ astacus.egg-info/
 
 
 .PHONY: build-dep-fedora
--- a/astacus/common/rohmustorage.py
+++ b/astacus/common/rohmustorage.py
@@ -66,6 +66,20 @@ class RohmuCompression(RohmuModel):
     level: int = 0
     # threads: int = 0
 
+    @property
+    def algorithm_value(self) -> str | None:
+        # The parent model is configured with `use_enum_values=True`,
+        # therefore the algorithm attribute can be string typed.
+        match self.algorithm:
+            case RohmuCompressionType():
+                return self.algorithm.value
+            case str():
+                return self.algorithm  # type: ignore[unreachable]
+            case None:
+                return None
+            case _:
+                raise TypeError(f"Expected compression algorithm, got {self.algorithm!r}")
+
 
 class RohmuConfig(RohmuModel):
     temporary_directory: str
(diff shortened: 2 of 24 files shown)


Makefile
```diff
@@ -11,19 +11,16 @@ GENERATED_PROTOBUFS = $(patsubst %.proto,%_pb2.py,$(PROTOBUFS))
...
-DNF_INSTALL = sudo dnf install -y
-
 ZOOKEEPER_VERSION = 3.8.4
 ZOOKEEPER_HASH = 284cb4675adb64794c63d95bf202d265cebddc0cda86ac86fb0ede8049de9187
 
 default: $(GENERATED)
 
 venv: default
-	$(PYTHON) -m venv venv && source venv/bin/activate && pip install -U pip && pip install . ".[dev]"
+	python3 -m venv .venv && source .venv/bin/activate && pip install -U pip && pip install . ".[dev]"
```
kmichel-aiven, line 20: Why rename the venv ?

tests/utils.py
```diff
@@ -43,7 +42,7 @@
 -----END PRIVATE KEY-----"""
 
 
-def create_rohmu_config(tmpdir: py.path.local, *, compression: bool = True, encryption: bool = True) -> RohmuConfig:
+def create_rohmu_config(tmpdir: os.PathLike, *, compression: bool = True, encryption: bool = True) -> RohmuConfig:
```
kmichel-aiven, line 45: tmpdir (the python fixture) is really a py.path.local, which has a rather different API from Path.

However, it looks like we're actually passing a path from `tmp_path` (the fixture), which is really a Path. I think it would be clearer to rename all the `tmpdir` to `tmp_path` and type everything as `Path`.

astacus/common/rohmustorage.py
```diff
@@ -148,20 +163,22 @@ def _loaded_public_key_lookup(self, key_id: str) -> RSAPublicKey:
...
+            wrapped_file = CompressionStream(wrapped_file, compression.algorithm_value, compression.level)
         if encryption_key_id:
             metadata.encryption_key_id = encryption_key_id
             rsa_public_key = self._loaded_public_key_lookup(encryption_key_id)
             wrapped_file = EncryptorStream(wrapped_file, rsa_public_key)
         rohmu_metadata = metadata.dict(exclude_defaults=True, by_alias=True)
-        self.storage.store_file_object(key, wrapped_file, metadata=rohmu_metadata)
+        # BaseTransfer.store_file_object is compatible with file descriptors of type IO[bytes] and Stream
+        # However, the BinaryIO parameter type is too strict for the union type we want to pass here.
+        self.storage.store_file_object(key, wrapped_file, metadata=rohmu_metadata)  # type: ignore[arg-type]
```
kmichel-aiven, line 181: It's a bit risky to silence type exactly at the boundary to an external library. Especially since we know `store_file_object` can have multiple implementation for each storage providers and where having each party uphold the promise of the interface type would be important for making sure this will work, without testing the product of all library users with all storage providers.

This looks like a bug in rohmu then, it shouldn't take a BinaryIO and use a type that only requires the "read" part of the interface ?

(That's not blocking, since the type was completely unchecked before this PR, so it's not worse than before.)

tests/utils.py
aris-aiven: Didn't want to do the refactoring in this PR, but I can.

astacus/common/rohmustorage.py
aris-aiven: Rohmu's storage interface needs to be fixed. As far as I can tell, the `fd` should abide by a protocol that implements `seek`, `tell` and `read`. Not sure how it type checks at the moment, since `Stream` is not a `BinaryIO` subclass.

Makefile
aris-aiven: Would you prefer the name wasn't renamed?

tests/utils.py
aris-aiven: It's not a refactoring just a rename, done.

Makefile
kmichel-aiven: Yes, I don't think this should be renamed (and the rename doesn't match the make target anymore).

aris-aiven:
Also pinned pydantic to v1 for the time being, as pip installs v2 by default.

Removing some more artifacts on the clean target.

kmichel-aiven (approved):

PR #215, opened 2024-04-16
Optionally ignore imbalanced AZs on `MapNode` step

aris-aiven:
When restoring a cluster, the new nodes AZ distribution may not match the one stored in the backup manifest. In this case, the `MapNode` fails even though the new nodes may actually be better distributed now.

The change allows mapping new nodes to their backup index in the presence of imperfect match between node AZs and backup AZs. A new exception is added to identify the imbalance error and catch it.

The option to the step defaults to false so as to retain backwards compatibility with the previous behaviour.

Ideally, a better solution would be to consider an AZ imbalance metric relative to some target replication factor, instead of naively remapping the backup indices. However, let's say this is good enough for the moment.

### DIFF
--- a/astacus/common/exceptions.py
+++ b/astacus/common/exceptions.py
@@ -24,6 +24,10 @@ class InsufficientAZsException(PermanentException):
     pass
 
 
+class ImbalancedAZsException(PermanentException):
+    pass
+
+
 class NotFoundException(PermanentException):
     pass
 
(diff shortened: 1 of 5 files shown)


aris-aiven:
This is not a good solution, and the problem shouldn't persist the both backup and new node AZ distributions are balanced.

PR #216, opened 2024-04-21
Use thread pool for DeleteDanglingObjectStorageFilesStep

kmichel-aiven:
CPU intensive steps running in the main thread block all async operations.

This include other concurrent requests, leaving astacus completely unresponsive.

Start moving some steps entirely in a thread pool to avoid that.

We were already using the thread pool a lot anyway, since our object storage library
is sync, but now instead of entering/leaving the pool for every small operation, we
run an entire step as a single task.

[DDB-951]


astacus/coordinator/plugins/base.py
```diff
@@ -17,6 +17,7 @@
 from astacus.coordinator.manifest import download_backup_manifest, download_backup_min_manifest
 from collections import Counter
 from collections.abc import Sequence, Set
+from starlette.concurrency import run_in_threadpool
```
alanfranz, line 20: I see no starlette dependency neither in astacus nor in aiven-core. I can imagine there's a transitive dependency, but if we start using it directly we should declare a direct dependency.

alanfranz (commented):
I need to take a further look, there're a few changes - I should be able to do that in a couple of hours. BTW:

>  since our object storage library is sync

This seems quite a shortcoming and I wonder whether this should be addressed somehow.

kmichel-aiven:
> > since our object storage library is sync
> 
> This seems quite a shortcoming and I wonder whether this should be addressed somehow.

It's also true of the zookeeper client, cassandra client, etc. Approximately everything except astacus client/server HTTP is sync code wrapped in threads and async awaited.

Given than "doing CPU-intensive work" happens in many place and is hard to diagnose (problems only happens with large backups), I'm leaning towards addressing it by progressively moving everything to threads and getting rid of async in all of astacus, not by trying to find async libraries for everything. We don't really need the theoretical efficiency of an astacus able to answer to thousand of concurrent queries.

(By "moving to threads", I mean "one query per thread" + "a few pools for things that are worth parallelizing and currently handled with asyncio.gather", not total chaos and tons of locks.)

astacus/coordinator/plugins/base.py
alanfranz: You're right it was already there, it wasn't so urgent.

alanfranz:
>  I'm leaning towards addressing it by progressively moving everything to threads and getting rid of async in all of astacus,

+1 on this. Async applications are usually efficient when everything is async, with only the occasional reliance on threads. If it's "oh my God it's full of threads" it's probably just easier to do everything in sync, rather than wasting time on making the two "colors" interoperate.

alanfranz (approved):
LGTM

PR #219, opened 2024-04-30
Use different dev dependencies for different fedora versions

nicois:
It is not sufficient to use one set of dev dependencies. In particular, pytest-asyncio has introduced a breaking change in f39.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,7 +32,7 @@ jobs:
           cache: "pip"
 
       - name: Install requirements
-        run: pip install -e '.[cassandra,dev,f${{ matrix.fedora-version }}]'
+        run: pip install -e '.[cassandra,f${{ matrix.fedora-version }}-dev,f${{ matrix.fedora-version }}]'
 
       - name: Execute lints and tests
         run: make tests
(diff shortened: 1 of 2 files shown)


tests/integration/conftest.py
```diff
@@ -147,7 +137,7 @@ def fixture_ports() -> Ports:
     return Ports()
 
 
-@pytest.fixture(scope="module", name="zookeeper")
+@pytest.fixture(name="zookeeper")
```
kmichel-aiven, line 140: We really can't drop all module scopes fixtures, the tests are already quite slow and it's not viable to move everything to function scope.

(Don't look at the GHA test times, many tests don't run in that context.)

Thankfully we already have a solution used in core, I've added a PR to do the same in astacus: https://github.com/Aiven-Open/astacus/pull/220

kmichel-aiven (changes requested):

tests/integration/conftest.py
nicois: oh good, I'd far prefer to move away from `pytest-asyncio`. I'll amend this PR to not include those changes, and I'll review the versions to make sure they include only the versions included there.

PR #225, opened 2024-06-25
clickhouse: backup user defined functions [DDB-1034]

joelynch:


astacus/coordinator/plugins/clickhouse/README.md
```diff
@@ -43,6 +43,7 @@
   },
   "replicated_access_zookeeper_path": "/clickhouse/access",
   "replicated_databases_zookeeper_path": "/clickhouse/databases",
+  "replicated_user_defined_zookeeper_path": "/clickhouse/udfs/",
```
kmichel-aiven, line 46: Why this one has a trailing slash but not the others ?

astacus/coordinator/plugins/clickhouse/manifest.py
```diff
@@ -25,15 +26,26 @@ class AccessEntity(AstacusModel):
     attach_query: bytes
 
     @classmethod
-    def from_plugin_data(cls, data: Mapping[str, Any]) -> "AccessEntity":
-        return AccessEntity(
+    def from_plugin_data(cls, data: Mapping[str, Any]) -> Self:
+        return cls(
```
kmichel-aiven, line 30: nit. Unrelated change, this should be in a separate commit (same with all similar changes).

astacus/coordinator/plugins/clickhouse/steps.py
```diff
@@ -151,6 +153,28 @@ async def run_step(self, cluster: Cluster, context: StepsContext) -> Sequence[Ac
...
+    replicated_user_defined_zookeeper_path: str | None
+
+    async def run_step(self, cluster: Cluster, context: StepsContext) -> Sequence[UDF]:
+        if self.replicated_user_defined_zookeeper_path is None:
+            return []
+
+        udfs: list[UDF] = []
+        async with self.zookeeper_client.connect() as connection:
+            change_watch = ChangeWatch()
+            children = await connection.get_children(self.replicated_user_defined_zookeeper_path, watch=change_watch)
```
kmichel-aiven, line 168: Can `/clickhouse/udfs` be missing if the user has not created any function yet ? (This should probably handle `NoNodeError` if that's not the case)

astacus/coordinator/plugins/clickhouse/steps.py
```diff
@@ -151,6 +153,28 @@ async def run_step(self, cluster: Cluster, context: StepsContext) -> Sequence[Ac
...
+            for child in children:
+                udf_path = os.path.join(self.replicated_user_defined_zookeeper_path, child)
```
kmichel-aiven, line 170: nit. The other places where we handle ZooKeeper paths are intentionally not using `os.path` because ZooKeeper znode paths are not filesystem paths.
(On linux, it _should_ be fine, since I don't think it's possible to create a child node with '/' in its name, so just a nit because it's odd to have inconsistent styles in the same file, shouldn't be a security issue.)

astacus/coordinator/plugins/clickhouse/steps.py
```diff
@@ -627,6 +653,22 @@ async def run_step(self, cluster: Cluster, context: StepsContext) -> None:
...
+
+    async def run_step(self, cluster: Cluster, context: StepsContext) -> None:
+        if self.replicated_user_defined_zookeeper_path is None:
+            return
+
+        clickhouse_manifest = context.get_result(ClickHouseManifestStep)
+        async with self.zookeeper_client.connect() as connection:
+            for udf in clickhouse_manifest.udfs:
+                path = os.path.join(self.replicated_user_defined_zookeeper_path, udf.path)
+                await connection.create(path, udf.value)
```
kmichel-aiven, line 669: The tests using `wait_for_udf` suggests that when the restore is done, it's not actually done, it should be done.

The step itself should wait for the functions to be available before finishing the step.

astacus/coordinator/plugins/clickhouse/steps.py
```diff
@@ -627,6 +653,22 @@ async def run_step(self, cluster: Cluster, context: StepsContext) -> None:
...
+    replicated_user_defined_zookeeper_path: str | None
+
+    async def run_step(self, cluster: Cluster, context: StepsContext) -> None:
+        if self.replicated_user_defined_zookeeper_path is None:
+            return
+
+        clickhouse_manifest = context.get_result(ClickHouseManifestStep)
+        async with self.zookeeper_client.connect() as connection:
+            for udf in clickhouse_manifest.udfs:
+                path = os.path.join(self.replicated_user_defined_zookeeper_path, udf.path)
```
kmichel-aiven, line 668: This probably needs a `connection.try_create(self.replicated_user_defined_zookeeper_path)` to make sure the entire path containing all functions exists ?

kmichel-aiven (dismissed):
A general thing: everything, including the commit message and the PR, just says "UDFs" without ever saying what is an UDF. I think it would be clearer if everything was named fully: `UserDefinedFunctions` (and `user_defined_functions` in the manifest).

astacus/coordinator/plugins/clickhouse/steps.py
joelynch: I think it is more about allowing for people to have a trailing slash on their zookeeper paths.

astacus/coordinator/plugins/clickhouse/steps.py
joelynch: Unneeded

https://github.com/Aiven-Open/astacus/blob/joelynch/backup-zk-udfs/astacus/coordinator/plugins/zookeeper.py#L89

astacus/coordinator/plugins/clickhouse/steps.py
joelynch: Good point

astacus/coordinator/plugins/clickhouse/steps.py
kmichel-aiven: Ah right, I forgot that outside of a transaction it's possible/easy to auto-create the entire path :+1:

tests/unit/coordinator/plugins/clickhouse/test_steps.py
```diff
@@ -131,6 +134,15 @@
     )
 ]
 
+SAMPLE_UserDefinedFunctionS = [
```
kmichel-aiven, line 137: Constant name is wrong.

tests/unit/coordinator/plugins/clickhouse/test_manifest.py
```diff
@@ -46,6 +48,24 @@
...
 }
+SAMPLE_UserDefinedFunctionS = [
+    UserDefinedFunction(
+        path="user_defined_function_1.sql", create_query=b"CREATE FUNCTION user_defined_function_1 AS (x) -> x + 1;\n"
+    ),
+    UserDefinedFunction(
+        path="user_defined_function_2.sql", create_query=b"CREATE FUNCTION user_defined_function_2 AS (x) -> x + 2;\n"
+    ),
+]
+SERIALIZED_UserDefinedFunctionS = [
```
kmichel-aiven, line 59: Constant names are wrong.

astacus/coordinator/plugins/clickhouse/steps.py
```diff
@@ -627,6 +655,51 @@ async def run_step(self, cluster: Cluster, context: StepsContext) -> None:
...
+        create_queries: list[str] = []
+        for function in clickhouse_manifest.user_defined_functions:
+            create_queries.append(escape_sql_string(function.create_query.removesuffix(b"\n")))
+        create_queries_str = ", ".join(create_queries)
+        query = f"""\
+            SELECT count(*)
+            FROM system.functions
+            WHERE create_query IN ({create_queries_str})
+            AND origin = 'SQLUserDefined'
+        """.encode()
```
kmichel-aiven, line 687: After how many function definitions does that fail ? (I assume there is a limit on the size of a query)

astacus/coordinator/plugins/clickhouse/steps.py
```diff
@@ -151,6 +153,30 @@ async def run_step(self, cluster: Cluster, context: StepsContext) -> Sequence[Ac
...
+
+    async def run_step(self, cluster: Cluster, context: StepsContext) -> Sequence[UserDefinedFunction]:
+        if self.replicated_user_defined_zookeeper_path is None:
+            return []
+
+        user_defined_functions: list[UserDefinedFunction] = []
+        async with self.zookeeper_client.connect() as connection:
+            change_watch = ChangeWatch()
+            # In case the node does not exist, create it now.
+            await connection.try_create(self.replicated_user_defined_zookeeper_path, b"")
```
kmichel-aiven, line 169: I don't think a Retrieve function should create something, it should only read the system under backup.

astacus/coordinator/plugins/clickhouse/steps.py
```diff
@@ -627,6 +655,51 @@ async def run_step(self, cluster: Cluster, context: StepsContext) -> None:
...
+        create_queries_str = ", ".join(create_queries)
+        query = f"""\
+            SELECT count(*)
+            FROM system.functions
+            WHERE create_query IN ({create_queries_str})
+            AND origin = 'SQLUserDefined'
+        """.encode()
+
+        async def wait_for_functions(client: ClickHouseClient) -> None:
+            for _ in range(10):
```
kmichel-aiven, line 690: All other step timeouts are configurable, this one should be configurable too (that has been helpful during incidents).

I think you can bump it to a higher value too, like 60sec, it would be annoying to fail a restore just because there are many functions or ZK/ClickHouse struggled temporarily for ~10sec.

kmichel-aiven (dismissed):

astacus/coordinator/plugins/clickhouse/steps.py
joelynch: Can do

astacus/coordinator/plugins/clickhouse/steps.py
kmichel-aiven: Counting before/after will probably fail when doing retries (the entire restore procedure can fail and retry).

I think just counting the total is good enough :+1:

tests/unit/coordinator/plugins/clickhouse/test_manifest.py
```diff
@@ -46,6 +48,24 @@
...
 }
+SAMPLE_USER_DEFINED_FUNCTIONS = [
+    UserDefinedFunction(
+        path="user_defined_function_1.sql", create_query=b"CREATE FUNCTION user_defined_function_1 AS (x) -> x + 1;\n"
+    ),
+    UserDefinedFunction(
+        path="user_defined_function_2.sql", create_query=b"CREATE FUNCTION user_defined_function_2 AS (x) -> x + 2;\n"
+    ),
+]
+SERIALIZED_UserDefinedFunctionS = [
```
joelynch, line 59: Oh I missed this one again

astacus/coordinator/plugins/clickhouse/steps.py
```diff
@@ -627,6 +679,39 @@ async def run_step(self, cluster: Cluster, context: StepsContext) -> None:
...
+            return
+
+        clickhouse_manifest = context.get_result(ClickHouseManifestStep)
+        if not clickhouse_manifest.user_defined_functions:
+            return
+
+        async with self.zookeeper_client.connect() as connection:
+            for user_defined_function in clickhouse_manifest.user_defined_functions:
+                path = os.path.join(self.replicated_user_defined_zookeeper_path, user_defined_function.path)
+                await connection.create(path, user_defined_function.create_query)
```
kmichel-aiven, line 700: Sorry I missed that earlier.

There is something in `RestoreAccessEntitiesStep` that catches `NodeExistsError` in case its the second attempt at restoring (and the tests cover re-restoring after an initial failure), I think it should also be there. (The current code should be failing when running the full-backup-restore tests in `test_plugin.py` in your local env.)

astacus/coordinator/plugins/clickhouse/steps.py
```diff
@@ -435,6 +466,27 @@ async def run_on_every_node(
     await asyncio.gather(*[gather_limited(per_node_concurrency_limit, fn(client)) for client in clients])
 
 
+async def wait_for_condition_on_every_node(
```
kmichel-aiven, line 469: That's a good thing to add :+1: 

Can you write a specific test for the function itself ?

astacus/coordinator/plugins/clickhouse/steps.py
```diff
@@ -435,6 +466,27 @@ async def run_on_every_node(
...
+    clients: Iterable[ClickHouseClient],
+    condition: Callable[[ClickHouseClient], Awaitable[bool]],
+    description: str,
+    timeout_seconds: float,
+    recheck_every_seconds: float = 1.0,
+) -> None:
+    loop = asyncio.get_event_loop()
+
+    async def wait_for_condition(client: ClickHouseClient) -> None:
+        start_time = loop.time()
```
kmichel-aiven, line 479: It's unclear to me (and from reading the doc), is this time moving forward when another task in running ?

kmichel-aiven (changes requested):

astacus/coordinator/plugins/clickhouse/steps.py
joelynch: I just had a look at the implementation 😆  https://github.com/python/cpython/blob/main/Lib/asyncio/base_events.py#L775

I think I will cut out the middle man

kmichel-aiven (approved):

PR #226, opened 2024-07-02
fix missing arg to logger

sjamgade:


astacus/coordinator/plugins/base.py
```diff
@@ -445,7 +445,7 @@ async def run_step(self, cluster: Cluster, context: StepsContext) -> None:
             # Since deltas can be uploaded from a different set of nodes than the base backup,
             # explicitly re-match current nodes to delta manifest nodes on each restore.
             if self.partial_restore_nodes and not self.contains_partial_restore_hostnames(delta_manifest):
-                logger.info("Skipped %s, because no data from it maps to partial restore nodes")
+                logger.info("Skipped %s, because no data from it maps to partial restore nodes", delta_manifest)
```
kmichel-aiven, line 448: We probably should just print `delta_name` instead of the entire `delta_manifest` ?

kmichel-aiven (changes requested):

astacus/coordinator/plugins/base.py
sjamgade: yes, the manifest object seems pretty big, but delta_name does not tell much about the manifest, or does it ?

kmichel-aiven: It does not tell much, if you want to be the most helpful you probably need to extract all the data used by `contains_partial_restore_hostnames`: `[pr.backup_hostname for pr in self.partial_restore_nodes]` and `[sr.hostname in partial_restore_hostnames for sr in manifest.snapshot_results]`.

sjamgade: went with `delta_manifest.start` it tells something more individually identifiable

kmichel-aiven (approved):

PR #227, opened 2024-07-18
Close transfer objects when we're done using them [DDB-1160]

kmichel-aiven:
This avoids leaking sockets held in memory by some of the rohmu drivers.


astacus/common/rohmustorage.py
```diff
@@ -238,17 +241,3 @@ def upload_json_bytes(self, name: str, data: bytes | mmap.mmap) -> bool:
             data.seek(0)
             self._upload_key_from_file(key, data, len(data))
         return True
-
-
-class MultiRohmuStorage(MultiStorage[RohmuStorage]):
```
joelynch, line 243: At some point we need to do a deeper refactor of these storages IMO.

joelynch (approved):
Looks good

PR #229, opened 2024-08-22
astacus: track remaining restored object files

tilman-aiven:
Track the progress on the remaining object storage files that needs to be copied during `RestoreObjectStorageFilesStep`.
Expose as gauge `astacus_copy_tiered_object_storage_files_remaining_value`.

Syncing required with https://github.com/Aiven-Open/astacus/pull/228

Requires rohmu 2.7.0

- https://github.com/Aiven-Open/rohmu/pull/192

### DIFF
--- a/astacus/coordinator/plugins/clickhouse/object_storage.py
+++ b/astacus/coordinator/plugins/clickhouse/object_storage.py
@@ -13,6 +13,7 @@ from typing import Any
 import contextlib
 import dataclasses
 import datetime
+import functools
 import logging
 import rohmu
 import tempfile
@@ -80,30 +81,41 @@ class ThreadSafeRohmuStorage(ObjectStorage):
             raise NotImplementedError("Copying items is only supported from another ThreadSafeRohmuStorage")
         with source.get_storage() as source_storage:
             with self.get_storage() as target_storage:
-                self._copy_items_between(keys, source_storage=source_storage, target_storage=target_storage, stats=stats)
-
-    def _copy_items_between(
-        self,
-        keys: Sequence[str],
-        *,
-        source_storage: BaseTransfer[Any],
-        target_storage: BaseTransfer[Any],
-        stats: StatsClient | None,
-    ) -> None:
-        match self.copy_config:
-            case DirectCopyConfig():
-                logger.info("Copying the keys using the cloud APIs")
-                target_storage.copy_files_from(source=source_storage, keys=keys)
-            case LocalCopyConfig():
-                logger.info("Copying the keys by downloading from source/uploading to target")
-                _copy_via_local_filesystem(
-                    keys, source=source_storage, target=target_storage, copy_config=self.copy_config, stats=stats
+                copy_items_between(
+                    keys,
+                    copy_config=self.copy_config,
+                    source_storage=source_storage,
+                    target_storage=target_storage,
+                    stats=stats,
                 )
 
     @contextlib.contextmanager
     def get_storage(self) -> Iterator[BaseTransfer[Any]]:
         with self._storage_lock:
             yield self._storage
+
+
+def copy_items_between(
+    keys: Sequence[str],
+    *,
+    copy_config: DirectCopyConfig | LocalCopyConfig,
+    source_storage: BaseTransfer[Any],
+    target_storage: BaseTransfer[Any],
+    stats: StatsClient | None,
+) -> None:
+    total_keys = len(keys)
+    match copy_config:
+        case DirectCopyConfig():
+            logger.info("Copying %i keys using the cloud APIs", total_keys)
+            emit_direct_copy_progress = (
+                functools.partial(emit_copy_progress_metric, stats, copy_config.method) if stats else None
+            )
+            target_storage.copy_files_from(source=source_storage, keys=keys, progress_fn=emit_direct_copy_progress)
+        case LocalCopyConfig():
+            logger.info("Copying %i keys by downloading from source/uploading to target", total_keys)
+            _copy_via_local_filesystem(
+                keys, source=source_storage, target=target_storage, copy_config=copy_config, stats=stats
+            )
 
 
 def _copy_via_local_filesystem(
@@ -115,13 +127,19 @@ def _copy_via_local_filesystem(
     stats: StatsClient | None,
 ) -> None:
     keys_to_copy = len(keys)
-    for keys_copied, key in enumerate(keys):
+    for keys_copied, key in enumerate(keys, start=1):
         with tempfile.TemporaryFile(dir=copy_config.temporary_directory) as temp_file:
             metadata = source.get_contents_to_fileobj(key, temp_file)
             target.store_file_object(key, temp_file, metadata)
         if stats:
-            stats.gauge(
-                "astacus_restore_clickhouse_tiered_storage_keys_remaining",
-                keys_to_copy - keys_copied,
-                tags={"copy_method": copy_config.method},
+            emit_copy_progress_metric(
+                stats=stats, copy_method=copy_config.method, completed_files=keys_copied, total_files=keys_to_copy
             )
+
+
+def emit_copy_progress_metric(stats: StatsClient, copy_method: str, completed_files: int, total_files: int) -> None:
+    stats.gauge(
+        "astacus_copy_tiered_object_storage_files_remaining",
+        total_files - completed_files,
+        tags={"copy_method": copy_method},
+    )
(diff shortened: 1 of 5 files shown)


astacus/coordinator/plugins/clickhouse/object_storage.py
```diff
@@ -91,12 +91,21 @@ def _copy_items_between(
...
+                    keys=keys,
+                    progress_fn=lambda completed_files, total_files: _handle_copy_progress_fn(
+                        completed_files, total_files, self.copy_config.method, stats
+                    )
+                    if stats
+                    else None,
+                )
             case LocalCopyConfig():
-                logger.info("Copying the keys by downloading from source/uploading to target")
+                logger.info("Copying %r keys by downloading from source/uploading to target", total_keys)
```
tilman-aiven, line 108: Since metrics are not a reliable data source, we add in addition the initial amount of keys for debugging purpose.

pyproject.toml
```diff
@@ -30,7 +30,7 @@ dependencies = [
     "protobuf < 3.21",
     "pydantic < 2",
     "pyyaml",
-    "rohmu >= 2.5.0",
+    "rohmu >= 2.7.0",
```
tilman-aiven, line 33: @jadkik 
this includes `2.6.0`, https://github.com/Aiven-Open/rohmu/releases/tag/releases%2F2.6.0

astacus/coordinator/plugins/clickhouse/object_storage.py
```diff
@@ -116,16 +125,20 @@ def _copy_via_local_filesystem(
...
         if stats:
-            stats.gauge(
-                "astacus_restore_clickhouse_tiered_storage_keys_remaining",
-                keys_to_copy - keys_copied,
-                tags={"copy_method": copy_config.method},
-            )
+            _handle_copy_progress_fn(keys_copied, keys_to_copy, copy_config.method, stats)
+
+
+def _handle_copy_progress_fn(completed_files: int, total_files: int, copy_method: str, stats: StatsClient) -> None:
```
aris-aiven, line 136: The stats client is usually the first arg. We also deviate from the typical style of systematically using underscores for helper functions.

aris-aiven: In this case, please make the args following the `stats` argument keywork only. IDEs usually complete the arg name, but reading `_handle_copy_progress_fn(stats, 1, 2, "mymethod")` out of context isn't very readable compared to `_handle_copy_progress_fn(stats, completed_files=1, total_files=2, copy_method="mymethod")`

tilman-aiven: Thanks for the advise

- added * after `stats`
- moved stats before additional arguments

```python
_handle_copy_progress_fn(stats: StatsClient, *, completed_files: int, total_files: int, copy_method: str) -> None:
```

Expecting
```python
_handle_copy_progress_fn(stats, completed_files=1, total_files=2, copy_method="mymethod")
```

astacus/coordinator/plugins/clickhouse/object_storage.py
```diff
@@ -91,12 +91,24 @@ def _copy_items_between(
         target_storage: BaseTransfer[Any],
         stats: StatsClient | None,
     ) -> None:
+        total_keys = len(keys)
         match self.copy_config:
             case DirectCopyConfig():
-                logger.info("Copying the keys using the cloud APIs")
-                target_storage.copy_files_from(source=source_storage, keys=keys)
+                logger.info("Copying %r keys using the cloud APIs", total_keys)
```
aris-aiven, line 97: The correct string formatting keyword here is `%i` for integers.. sorry I didn't spot this earlier

astacus/coordinator/plugins/clickhouse/object_storage.py
aris-aiven: That's cool! The keyword should be `%i` instead of `%r`

aris-aiven (changes requested):
Another nit hehe

astacus/coordinator/plugins/clickhouse/object_storage.py
tilman-aiven: Adressed. 
Looking into our code base, we have not used once `%i` for integers.
I read that `repr` is the optimal way to debug values. However since its definitely `int`, this should be fine 👍

astacus/coordinator/plugins/clickhouse/object_storage.py
```diff
@@ -91,12 +91,24 @@ def _copy_items_between(
...
+        total_keys = len(keys)
         match self.copy_config:
             case DirectCopyConfig():
-                logger.info("Copying the keys using the cloud APIs")
-                target_storage.copy_files_from(source=source_storage, keys=keys)
+                logger.info("Copying %i keys using the cloud APIs", total_keys)
+                target_storage.copy_files_from(
+                    source=source_storage,
+                    keys=keys,
+                    progress_fn=lambda completed_files, total_files: _handle_copy_progress_fn(
```
dmitry-potepalov, line 101: discussed in more detail, but basically I want some changes to the code structure:
* rename the handler to `emit_.._metric` to declare intent
* hide the details of direct copy into a function similar to _copy_via_local to keep the match statement clear and forgettable
* maybe use functools.partial instead of lambda to bind method and stats

dmitry-potepalov (dismissed):
Logic OK, but the code needs to be cleaned up a bit, see comment

astacus/coordinator/plugins/clickhouse/object_storage.py
```diff
@@ -107,6 +96,29 @@ def get_storage(self) -> Iterator[BaseTransfer[Any]]:
             yield self._storage
 
 
+def copy_items_between(
```
tilman-aiven, line 99: extracted method from class for testing purpose.
`test_steps` uses an abstraction `MemoryObjectStorage`, which does not reference `copy_items_between`.
Added a test in `test_copy_items_between_with_direct_copy_method` in order to verify `emit_direct_copy_progress` for `DirectCopyConfig`.

astacus/coordinator/plugins/clickhouse/object_storage.py
tilman-aiven: See https://github.com/Aiven-Open/astacus/pull/229#discussion_r1734481164

astacus/coordinator/plugins/clickhouse/object_storage.py
dmitry-potepalov: IMHO not really important to have it private, especially since it's tested now; let's have it public

astacus/coordinator/plugins/clickhouse/object_storage.py
```diff
@@ -159,6 +177,14 @@ def delete_item(self, key: str) -> None:
 
     def copy_items_from(self, source: "ObjectStorage", keys: Sequence[str], *, stats: StatsClient | None) -> None:
         keys_set = set(keys)
-        for source_item in source.list_items():
+        keys_to_copy = len(keys_set)
+        for keys_copied, source_item in enumerate(source.list_items(), start=1):
             if source_item.key in keys_set:
                 self.items[source_item.key] = source_item
+            if stats:
+                _emit_copy_progress_metric(
```
dmitry-potepalov, line 185: maybe just move the `MemoryObjectStorage` into `tests`? since it's not used in prod, emitting metrics from it does not add much value; but since it's near the prod code, one is tempted to add prod behavior to it

dmitry-potepalov (commented):
left some nits on tests, IMHO this can go in as-is, the comments can be addressed in a subsequent PR

tests/unit/coordinator/plugins/clickhouse/test_steps.py
```diff
@@ -1010,21 +1020,67 @@ async def test_restore_replica() -> None:
...
     context = StepsContext()
     context.set_result(ClickHouseManifestStep, SAMPLE_MANIFEST)
-    await step.run_step(cluster, context)
+
+    with patch.object(stats, "gauge") as mock_stats_gauge:
+        await step.run_step(cluster, context)
+
     assert target_object_storage.list_items() == object_storage_items
+    if with_stats:
+        assert len(clickhouse_source_object_storage_files) == mock_stats_gauge.call_count
```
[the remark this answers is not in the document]
tilman-aiven: Good idea.
However the assertion below is not safe in term of counting. 
The number of items in `clickhouse_source_object_storage_files` could change and at the same time the progress function does not cover "a last item".

astacus/coordinator/plugins/clickhouse/object_storage.py
tilman-aiven: Moved to `tests/unit/coordinator/plugins/clickhouse/object_storage.py` in a separate commit.

tilman-aiven:
@dmitry-potepalov thanks for feedback.
comments addressed.

dmitry-potepalov (approved):
LGTM

PR #231, opened 2024-08-28
cover local copy in restore integration test

dmitry-potepalov:
Run the restore integration suite both with direct and local copy.


aris-aiven:
Didn't get to it today, reviewing tomorrow.

kmichel-aiven (changes requested):
I don't think we should add combinatorial explosion of parameters of the entire backup/restore integration tests.

Instead, the two copy variants should be tested by directly testing `ThreadSafeRohmuStorage.copy_items_from`.

PR #232, opened 2024-08-29
clickhouse: flatten_nested no longer works for Array(Tuple) in 24.3

joelynch:
https://github.com/ClickHouse/ClickHouse/pull/56132.

Cache the output of get_clickhouse_version if we are going to call it a lot of times.  With this, all the Astacus tests pass for ClickHouse 24.3.


dmitry-potepalov (approved):
LGTM

PR #233, opened 2024-09-06
common: allow extra fields in AstacusModel

joelynch:
Without this it is very hard to add additional fields to any api endpoints, node or coordinator.


joelynch:
Maybe a better approach is to just use msgspec everywhere and be more consistent with this.

PR #234, opened 2024-09-06
clickhouse: add the total size on object storage to backup list [DDB-1222]

joelynch:


astacus/coordinator/list.py
```diff
@@ -39,6 +41,22 @@ def compute_deduplicated_snapshot_file_stats(manifest: ipc.BackupManifest) -> tu
     return num_files, total_size
 
 
+def get_object_storage_size_bytes(manifest: ipc.BackupManifest) -> int | None:
+    """Compute the total size of object storage files for the backup in bytes."""
+    if manifest.plugin == ipc.Plugin.clickhouse:
+        clickhouse_data = ClickHouseManifest.from_plugin_data(manifest.plugin_data)
```
dmitry-potepalov, line 47: isn't it expensive to parse?... would it be better to add a new field to the base manifest, set to 0 for all other service types and populate it for CH when creating the backup? e.g. add something a-la:

```
class BackupManifest(...
...
    # What did the upload return (mostly for statistics)
    upload_results: Sequence[SnapshotUploadResult]

    # How much bytes are we keeping in object storage
    tiered_storage_results: Sequence[TieredStorageResult]
```
to ipc.BackupManifest? Recalculating this value feels wrong, because manifests are supposed to be immutable.

dmitry-potepalov (changes requested):
not sure re-computing the value every time we list the backup is a good idea, open to debate - but I'll request changes for now

astacus/coordinator/list.py
joelynch: Yea agreed that parsing the plugin data is not great.  Although I have been thinking about [this](https://jcristharif.com/msgspec/supported-types.html#generic-types) to make it faster potentially.

Since `UploadManifestStep` is shared, what is the least awkward way to add this to the base manifest do you think?

joelynch: BTW confusingly `clickhouse_data.object_storage_files` will usually have length 1 (since it has one value per CH disk) so it is not doing much computation after reparsing the plugin data.

joelynch:
Okay I've redone it with an extra step `CollectTieredStorageResultsStep` to pull the object storage file sizes from `CollectObjectStorageFilesStep`.

tests/unit/coordinator/test_list.py
```diff
@@ -239,7 +245,7 @@ def test_api_list_deduplication(backup_manifest: BackupManifest, tmp_path: Path)
                         name="1",
                         start=datetime.datetime(2020, 1, 2, 3, 4, 5, 678, tzinfo=datetime.timezone.utc),
                         end=datetime.datetime(2020, 1, 2, 5, 6, 7, 891, tzinfo=datetime.timezone.utc),
-                        plugin=Plugin("clickhouse"),
```
dmitry-potepalov, line 242: is this change actually necessary, given the fact that we've moved things around and no longer require list to parse the plugin manifest?

dmitry-potepalov (approved):
LGTM

PR #236, opened 2024-09-12
ci: use ruff formatter

joelynch:
The ruff formatter is much faster than using black and isort.


dmitry-potepalov:
I'll check it out and see whether I run into any issues with my env

dmitry-potepalov (changes requested):
this breaks import sorting, because import sorting is a part of the ruff linter, not the formatter :( .

joelynch:
> this breaks import sorting, because import sorting is a part of the ruff linter, not the formatter :( .

Okay I have put isort back in since ruff linting is a lot of changes.  In a future PR I will do ruff linting.

dmitry-potepalov (approved):
LGTM

PR #239, opened 2024-09-20
clickhouse: don't log error when keeper_map path does not exist

joelynch:
This will always happen if there is no KeeperMap tables.


tilman-aiven (approved):
LGTM

PR #240, opened 2024-09-25
clickhouse: account for dropped KeeperMap tables

joelynch:
If the table was dropped the table path still exists but the data ZNode is removed.  This leads to an error in backup.

### DIFF
--- a/astacus/coordinator/plugins/clickhouse/steps.py
+++ b/astacus/coordinator/plugins/clickhouse/steps.py
@@ -204,7 +204,12 @@ class RetrieveKeeperMapTableDataStep(Step[Sequence[KeeperMapTable]]):
             for child in children:
                 keeper_map_table_path = os.path.join(self.keeper_map_path_prefix, child)
                 data_path = os.path.join(keeper_map_table_path, "data")
-                data = await connection.get_children_with_data(data_path)
+                try:
+                    data = await connection.get_children_with_data(data_path)
+                except NoNodeError:
+                    logger.info("ZNode %s is missing, table was dropped.  Skipping", data_path)
+                    continue
+
                 tables.append(
                     KeeperMapTable(
                         name=child,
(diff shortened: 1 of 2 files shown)


astacus/coordinator/plugins/clickhouse/steps.py
```diff
@@ -204,7 +204,12 @@ async def run_step(self, cluster: Cluster, context: StepsContext) -> Sequence[Ke
             for child in children:
                 keeper_map_table_path = os.path.join(self.keeper_map_path_prefix, child)
                 data_path = os.path.join(keeper_map_table_path, "data")
-                data = await connection.get_children_with_data(data_path)
+                try:
+                    data = await connection.get_children_with_data(data_path)
+                except NoNodeError:
+                    logger.warning("ZNode %s is missing, table was dropped.  Skipping", data_path)
```
Khatskevich, line 210: It it is just dropped data, probably it should be info or debut level log, nothing to worry about.

tests/integration/coordinator/plugins/clickhouse/test_plugin.py
```diff
@@ -316,21 +316,32 @@ async def setup_cluster_content(clients: Sequence[HttpClickHouseClient], use_nam
...
-            "waiting for keeper_map to be replicated",
-            timeout_seconds=5,
-        )
+async def add_keeper_map_tables(clients: Sequence[ClickHouseClient]) -> None:
+    # add and drop a table with KeeperMap engine.  This leaves some metadata in ZooKeeper that we want to ignore.
+    await clients[0].execute(
+        b"CREATE TABLE default.keeper_map_dropped(key UInt32, value String) "
+        b"ENGINE = KeeperMap('keeper_map_dropped') PRIMARY KEY key"
+    )
+    await clients[0].execute(b"DROP TABLE default.keeper_map_dropped")
```
Khatskevich, line 328: This test checks if deleted table does not cause problems, but it does not tell if the new `try except` kicks it there. I will try to delete keeper map table and see what is in the zookeeper. But It would be even better if there will be a test showing that exactly those lines are executed when the table was deleted. (Because there may be intersection with other cases, like empty table, or deleted table deletes metadata and the issue is not fixed)

Khatskevich (changes requested):

tests/integration/coordinator/plugins/clickhouse/test_plugin.py
joelynch: This modified test fails on main, indicating that the added code is what kicks in.

You can see [here](https://github.com/ClickHouse/ClickHouse/blob/24.3/src/Storages/StorageKeeperMap.cpp#L566) that all zk paths except the parent path `zk_root_path` are dropped.

joelynch: I will add an empty table as a test case as well for completeness :)

Khatskevich: Thanks. I have also double checked it

Khatskevich (approved):
Looks good

PR #243, opened 2024-10-02
Migrate mypy and ruff config to pyproject.toml

joelynch:
* Bump mypy version
* We don't use tox anymore.


aris-aiven:
Could you remove the following from `.github/workflows/test.yml` and eventually the related pins in the `pyproject.toml`?
```
        include:
          - fedora-version: "38"
            python-version: "3.11"
```
We might eventually replace it with the following (but not necessarily in this PR)
```
        include:
          - fedora-version: "40"
            python-version: "3.13"
```
This should make the pipeline green

pyproject.toml
```diff
@@ -148,3 +149,168 @@ source = [
...
+module = "google.protobuf"
+ignore_errors = true
+
+[[tool.mypy.overrides]]
+module = "astacus.common.cassandra,astacus.coordinator.cassandra,astacus.node.cassandra"
+disallow_incomplete_defs = true
+disallow_untyped_defs = true
+
+[tool.ruff]
+target-version = "py311"
```
aris-aiven, line 191: This we can target `py312` here (for a follow up)

aris-aiven (approved):
LGTM

PR #244, opened 2024-10-02
Remove FastAPI dependency

joelynch:
FastAPI does not suit our needs as our routes do not return pydantic models, they return MsgSpec Structs. Additionally, we do not make use of schema generation.  Instead, we can thinly wrap Starlette ourselves.


kmichel-aiven (changes requested):
I'm not sure keeping a decorator-based approach is really necessary, that's more code and complexity than just building the routers with direct calls. I don't think it's worth writing half a custom framework.

The other related aspect is that `starlette.py` isn't tested, so making it much smaller and directly testing it would be good.

PR #245, opened 2024-10-09
Allow all possible experimental indexes on restore

dmitry-potepalov:
Presently we don't know which experimental features/indexes were enabled during table creation, because those could have been enabled on a per-session/per-query basis and not captured in the create_table_query. To avoid failures on restore, just enable all experimental features when creating the tables. Should be relatively harmless, since it affects only the restoration session; also most of those flags do not affect tables that don't have experimental indexes.


astacus/coordinator/plugins/clickhouse/steps.py
```diff
@@ -631,11 +631,30 @@ def _create_dbs(client: ClickHouseClient) -> Iterator[Awaitable[None]]:
...
-            b"SET allow_experimental_object_type=true",
-            b"SET allow_suspicious_codecs=true",
+            b"SET allow_experimental_map_type=true",
             b"SET allow_suspicious_low_cardinality_types=true",
+            b"SET allow_suspicious_fixed_string_types=true",
+            b"SET allow_suspicious_indices=true",
+            b"SET allow_suspicious_codecs=true",
+            b"SET allow_hyperscan=true",
+            b"SET allow_simdjson=true",
+            b"SET allow_deprecated_syntax_for_merge_tree=true",
```
Khatskevich, line 657: Good work. The only thing I am concerned about is compatibility.
Previously the list was conservative and now it is quite wide.
It is possible that some very new option is not available on a running clickhouse, and will cause error like this
```
clickhouse01 :) set test=1

Received exception from server (version 24.8.4):
Code: 115. DB::Exception: Received from localhost:9000. DB::Exception: Unknown setting 'test'. (UNKNOWN_SETTING)
```
Which is extremely unpleasant during restoration.
I think it makes sense to just ignore those errors or check if the setting is available using `system.settings` table.

dmitry-potepalov: > unless we manually call /restore on a running node

and that's actually also an edge case for a future update of that list, since the old nodes don't have the extended list and the new nodes already have 23.8 LTS

Khatskevich: You are right, but I think it is easier to just add the safety guard and forget about possible problems:
1. may be someone is using this open source project)
2. it is actually better to add all the settings from the latest clickhouse, rather then currently used lts
3. situation may change in the future it is better to just be prepared
I really do not see a reason to not protect this place.

Khatskevich (approved):

PR #248, opened 2024-10-19
Migrate from pylint and flake8 to ruff

aris-aiven:
Update pre-commit hooks and transition to ruff.
This simplifies the linting workflow and improves its performance.


Makefile
```diff
@@ -69,15 +62,15 @@ unittest: $(GENERATED)
 test: lint copyright unittest
 
 .PHONY: isort
-isort:
+isort: $(GENERATED)
 	pre-commit run isort --all-files
```
joelynch, line 66: Do we still need isort?  Ruff sorts imports.

joelynch (changes requested):

Makefile
aris-aiven: Good idea, just removing isort b/c the current ruff config should cover sorting AFAICT

joelynch (approved):

PR #249, opened 2024-10-24
Cleanup ClickHouse integration conftest

aris-aiven:
The conditionals were valid for old setups. We can use named collections and ZooKeeper macro placeholders everywhere now.


dmitry-potepalov (approved):
LGTM

PR #251, opened 2024-10-25
Zero copy tests

joelynch:
* Set zero copy settings in tests.
* make sure object storage test table has data both on local and remote disk.

### DIFF
--- a/tests/integration/coordinator/plugins/clickhouse/conftest.py
+++ b/tests/integration/coordinator/plugins/clickhouse/conftest.py
@@ -26,11 +26,7 @@ from pathlib import Path
 from tests.conftest import CLICKHOUSE_PATH_OPTION, CLICKHOUSE_RESTORE_PATH_OPTION
 from tests.integration.conftest import get_command_path, Ports, run_process_and_wait_for_pattern, Service, ServiceCluster
 from tests.system.conftest import background_process, wait_url_up
-from tests.utils import (
-    CONSTANT_TEST_RSA_PRIVATE_KEY,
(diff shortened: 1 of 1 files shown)


tests/integration/coordinator/plugins/clickhouse/test_plugin.py
```diff
@@ -313,6 +313,10 @@ async def setup_cluster_content(clients: Sequence[HttpClickHouseClient], clickho
     await clients[0].execute(b"INSERT INTO default.in_object_storage VALUES (123, 'foo')")
     await clients[1].execute(b"INSERT INTO default.in_object_storage VALUES (456, 'bar')")
     await clients[2].execute(b"INSERT INTO default.in_object_storage VALUES (789, 'baz')")
+    # these should be routed to local disk
```
Khatskevich, line 316: Why would they be routed to the local disk?
Even if there is a reason, it is better to test. (check if `system.parts` has parts on all disks for example)

joelynch: I don't think this is actually needed to reproduce the bugs I was seeing.  Removed.

Khatskevich: I do not understand what is happening here now.
Commit says: "make sure object storage test table has data both on local and remote disk", but there are 0 new tests in this pr, and everything looks to work just as it was. Why are you now sure that the data is stored on both disks during tests?

Khatskevich (approved):

PR #252, opened 2024-10-29
ClickHouse: pause KeeperMap insertion momentarily during backup

aris-aiven:
The PR alters replicated users' privileges (through GRANT & REVOKE) during the backup. This ensures the KeeperMap tables' snapshot is consistent with the frozen table data.

[DDB-1237]

### DIFF
--- a/astacus/coordinator/coordinator.py
+++ b/astacus/coordinator/coordinator.py
@@ -271,8 +271,9 @@ class SteppedCoordinatorOp(LockedCoordinatorOp):
                 with self._progress_handler(cluster, step):
                     try:
                         r = await step.run_step(cluster, context)
-                    except (StepFailedError, WaitResultError) as e:
-                        logger.info("Step %s failed: %s", step, str(e))
+                    except (StepFailedError, WaitResultError) as exc:
+                        logger.info("Step %s failed: %s", step, str(exc))
+                        await step.handle_step_failure(cluster, context)
                         return False
             context.set_result(step.__class__, r)
         return True
--- a/astacus/coordinator/plugins/clickhouse/steps.py
+++ b/astacus/coordinator/plugins/clickhouse/steps.py
@@ -43,6 +43,7 @@ from astacus.coordinator.plugins.base import (
 from astacus.coordinator.plugins.zookeeper import ChangeWatch, NoNodeError, TransactionError, ZooKeeperClient
 from base64 import b64decode
 from collections.abc import Awaitable, Callable, Iterable, Iterator, Mapping, Sequence
+from kazoo.exceptions import ZookeeperError
 from typing import Any, cast, TypeVar
 
 import asyncio
@@ -224,7 +225,7 @@ class KeeperMapTablesReadOnlyStep(Step[None]):
             timeout_seconds=60,
         )
 
-    async def run_step(self, cluster: Cluster, context: StepsContext):
+    async def run_step(self, cluster: Cluster, context: StepsContext) -> None:
         _, tables = context.get_result(RetrieveDatabasesAndTablesStep)
         replicated_users_response = await self.clients[0].execute(
             b"SELECT base64Encode(name) FROM system.users WHERE storage = 'replicated' ORDER BY name"
@@ -242,6 +243,7 @@ class KeeperMapTablesReadOnlyStep(Step[None]):
 class RetrieveKeeperMapTableDataStep(Step[Sequence[KeeperMapTable]]):
     zookeeper_client: ZooKeeperClient
     keeper_map_path_prefix: str | None
+    clients: Sequence[ClickHouseClient]
 
     async def run_step(self, cluster: Cluster, context: StepsContext) -> Sequence[KeeperMapTable]:
         if self.keeper_map_path_prefix is None:
@@ -254,6 +256,8 @@ class RetrieveKeeperMapTableDataStep(Step[Sequence[KeeperMapTable]]):
             except NoNodeError:
                 # The path doesn't exist, no keeper map tables to retrieve
                 return []
+            except ZookeeperError as e:
+                raise StepFailedError("Failed to retrieve KeeperMap tables") from e
 
             tables = []
             for child in children:
@@ -262,8 +266,10 @@ class RetrieveKeeperMapTableDataStep(Step[Sequence[KeeperMapTable]]):
                 try:
                     data = await connection.get_children_with_data(data_path)
                 except NoNodeError:
-                    logger.info("ZNode %s is missing, table was dropped.  Skipping", data_path)
+                    logger.info("ZNode %s is missing, table was dropped. Skipping", data_path)
                     continue
+                except ZookeeperError as e:
+                    raise StepFailedError("Failed to retrieve table data") from e
 
                 tables.append(
                     KeeperMapTable(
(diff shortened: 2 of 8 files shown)


tests/integration/coordinator/plugins/clickhouse/test_steps.py
```diff
@@ -99,3 +101,77 @@ async def test_retrieve_tables(ports: Ports, clickhouse_command: ClickHouseComma
...
+async def test_keeper_map_table_select_only_setting_modified(keeper_table_context: KeeperMapInfo) -> None:
+    steps_context, admin_client, user_client = keeper_table_context
+    read_only_step = KeeperMapTablesReadOnlyStep(clients=[admin_client], allow_writes=False)
+    # After the read-only step, the user should only be able to select from the table
+    await read_only_step.run_step(Cluster(nodes=[]), steps_context)
+    with pytest.raises(ClickHouseClientQueryError, match=".*ACCESS_DENIED.*"):
+        await user_client.execute(
+            b"INSERT INTO `keeperdata`.`keepertable` SETTINGS wait_for_async_insert=1  SELECT *, materialize(2) FROM numbers(3)"
+        )
+    with pytest.raises(ClickHouseClientQueryError, match=".*ACCESS_DENIED.*"):
```
Khatskevich, line 154: So, we are basically do it for the "ClickHouse connector for Kafka". I have not read it's code, but I suspect some new problems in the normal operation.
What does exactly once mean: every part is processed, every part is not processed 2 times.
Before this commit if there are not backup restore and errors during the operation it was exactly once.
Now, during backup some part will not be able to say that the work is done.
How exactly will it behave?
Because it theoretically could
1. fail instantly and someone else repeats the whole processing
2. be restarted and the processing is done again
If so, it is no better than it was, because previously it  was not happening during normal operation

astacus/coordinator/plugins/clickhouse/steps.py
```diff
@@ -179,6 +179,35 @@ async def run_step(self, cluster: Cluster, context: StepsContext) -> Sequence[Us
...
+        replicated_users_response = await self.clients[0].execute(
+            b"SELECT base64Encode(name) FROM system.users WHERE storage = 'replicated' ORDER BY name"
+        )
+        replicated_users_names = [b64decode(cast(str, user[0])) for user in replicated_users_response]
+        keeper_map_table_names = [table for table in tables if table.engine == "KeeperMap"]
+        privilege_altering_fun = self.get_grant_statement if self.allow_writes else self.get_revoke_statement
+        statements = [
+            privilege_altering_fun(table, escape_sql_identifier(user))
+            for table in keeper_map_table_names
+            for user in replicated_users_names
```
Khatskevich, line 206: 1. this looks a bit dangerous, because num_users*num_tables may blow up for some cases which we do not control.
2. and it might be not only just long, but will grow replication queue very fast
3. This replication queue not only should be stored, but also replayed by new replicas, repeating all those steps (as I understand)

astacus/coordinator/plugins/clickhouse/steps.py
```diff
@@ -179,6 +179,35 @@ async def run_step(self, cluster: Cluster, context: StepsContext) -> Sequence[Us
...
+        ]
+        await asyncio.gather(*(self.clients[0].execute(statement) for statement in statements))
```
Khatskevich, line 208: The effect is not immediate. Ideally, we should wait until this statement propagates to all the replicas

tests/integration/coordinator/plugins/clickhouse/test_steps.py
aris-aiven: The sink connector pauses writes to ClickHouse for these exceptions: https://github.com/ClickHouse/clickhouse-kafka-connect/blob/main/src/main/java/com/clickhouse/kafka/connect/util/Utils.java#L68

If the exception is retriable, the connector won't commit the offsets to the KeeperMap table, and won't try to write that part to ClickHouse until later. The design docs describes the state machine transitions: https://github.com/ClickHouse/clickhouse-kafka-connect/blob/main/docs/DESIGN.md

astacus/coordinator/plugins/clickhouse/steps.py
aris-aiven: I don't think that's a problem in practice, we usually have few users and little to no KeeperMap tables.

astacus/coordinator/plugins/clickhouse/steps.py
aris-aiven: Right, that makes sense, will update the logic to wait for the effect. Thanks!

tests/integration/coordinator/plugins/clickhouse/test_steps.py
aris-aiven: Critically, the `ACCESS_DENIED` raises exception code 497 which isn't included in the retriable exceptions list. So we need to change the approach a bit.

astacus/coordinator/plugins/clickhouse/steps.py
aris-aiven: I've played around with waiting for the correct grants to be in the `system.grants` table, or by using the `SHOW GRANTS FOR ALL`. Unfortunately that doesn't work for inherited grants, since they're computed on the fly and the table is "compressed". 
The other solution would be to have a sentinel value that we try to insert in the table, but then again it's bound to  succeed for the admin user?

aris-aiven: Anyhow there's no good way to wait for a GRANT statement to propagate. Let me know if the updated solution doesn't look to bad (I don't think we can aim for much better).

astacus/coordinator/coordinator.py
```diff
@@ -271,8 +271,9 @@ async def try_run(self, cluster: Cluster, context: StepsContext) -> bool:
                 with self._progress_handler(cluster, step):
                     try:
                         r = await step.run_step(cluster, context)
-                    except (StepFailedError, WaitResultError) as e:
-                        logger.info("Step %s failed: %s", step, str(e))
+                    except (StepFailedError, WaitResultError) as exc:
+                        logger.info("Step %s failed: %s", step, str(exc))
+                        await step.handle_step_failure(cluster, context)
```
joelynch, line 276: Oh I was thinking maybe more using a stack like

```python
failure_functions = []
for i, step in enumerate(self.steps, 1):
    failure_functions.append(step.handle_step_failure)
    ...
    try:
        ...
    except ...:
        for func in reversed(failure_functions):
            func(cluster, context)
```

But this is probably good enough

joelynch (approved):
Some suggestions that can be covered in a follow up

astacus/coordinator/coordinator.py
aris-aiven: Reading the code, if I understand correctly there are two scenarios:
1. If the failure is `Transient` then the step will be retried, so there's nothing to put on the stack yet. 
2. If the failure is permanent (`astacus.common.exceptions.PermanentException` subclass) then there won't be a next step. The exception clause in question sets the step to failed, and higher up the caller stack the whole `Op` is aborted (in this case the `BackupOp`).

Which case do you think could result in multiple steps needing to be unwound? 

> maybe the except needs to be more broad?
Looking at the usages for other `PermanentException` subclasses they're either raised during the restore steps or by the Rohmu storage subclasses (which would be caught when the server would be setup I think). Might be worth broadening `StepFailedError` to `PermanentException`..

PR #257, opened 2024-11-07
Only execute access rights change queries on a single node

aris-aiven:
For the read-only step, we're filtering by replicated users, whose grant information is stored in ZooKeeper.

This means that executing the query on multiple nodes is counterproductive. Instead of achieving faster information propagation, we're getting transaction errors from ZooKeeper. Any one of these can in turn fail the step.

Here's one such example:
```
Query failed: b'REVOKE INSERT, ALTER UPDATE, ALTER DELETE ON `default`.`keeper_map` FROM `alice`' on <host>:
status_code=500,
exception_code=999,
response=b'Code: 999. Coordination::Exception: Transaction failed: Op #0, path: /clickhouse/access/uuid/914a9caa-682f-5dce-4ed7-b0f16f564798. (KEEPER_EXCEPTION) (version 24.3.5.1)\n'
```

### DIFF
--- a/astacus/coordinator/plugins/clickhouse/steps.py
+++ b/astacus/coordinator/plugins/clickhouse/steps.py
@@ -190,14 +190,14 @@ class KeeperMapTablesReadOnlyStep(Step[None]):
         revoke_statement = (
             f"REVOKE INSERT, ALTER UPDATE, ALTER DELETE ON {table.escaped_sql_identifier} FROM {escaped_user_name}"
         )
-        await asyncio.gather(*(client.execute(revoke_statement.encode()) for client in self.clients))
+        await self.clients[0].execute(revoke_statement.encode())
 
     async def grant_write_on_table(self, table: Table, user_name: bytes) -> None:
         escaped_user_name = escape_sql_identifier(user_name)
         grant_statement = (
             f"GRANT INSERT, ALTER UPDATE, ALTER DELETE ON {table.escaped_sql_identifier} TO {escaped_user_name}"
         )
-        await asyncio.gather(*(client.execute(grant_statement.encode()) for client in self.clients))
+        await self.clients[0].execute(grant_statement.encode())
 
     async def run_step(self, cluster: Cluster, context: StepsContext) -> None:
         _, tables = context.get_result(RetrieveDatabasesAndTablesStep)


astacus/coordinator/plugins/clickhouse/steps.py
```diff
@@ -190,14 +190,14 @@ async def revoke_write_on_table(self, table: Table, user_name: bytes) -> None:
...
-        await asyncio.gather(*(client.execute(revoke_statement.encode()) for client in self.clients))
+        await self.clients[0].execute(revoke_statement.encode())
 
     async def grant_write_on_table(self, table: Table, user_name: bytes) -> None:
         escaped_user_name = escape_sql_identifier(user_name)
         grant_statement = (
             f"GRANT INSERT, ALTER UPDATE, ALTER DELETE ON {table.escaped_sql_identifier} TO {escaped_user_name}"
         )
-        await asyncio.gather(*(client.execute(grant_statement.encode()) for client in self.clients))
+        await self.clients[0].execute(grant_statement.encode())
```
Khatskevich, line 200: This change looks very controversial to me
1. Yes, in principle it is not necessary to change priileges multiple times, but we do not control how fast it will reach other servers, do we? Theoretically it might be delayed for unknown time which spoils all guarantees.
2. I really want to continue [this](https://aiven-io.slack.com/archives/C052RDBD1TP/p1730911269847339) discussion. So far I think it is better to just delete this step completely...

aris-aiven: I mean the cluster is in a single region, and zookeeper is strictly serializable, so the upper bound on the latency for the change to go through in ZK is <5ms.

Actually I don't think the deduplication log is relevant for the restore. The log is saved in ZooKeeper, and the restore process doesn't start from a ZK snapshot but from an empty state. So the block-level information is lost.

aris-aiven: Will test if in practice the argument I just provided stands, but the main branch is broken at the moment and I'd rather we fix it. If the step is irrelevant, then of course we'll drop it.

Khatskevich: Zookeeper only supports write operation serialization, but reads could propagate slowly.
They first go to the local zookeeper, and then to the clickhouse to be processed, as I understand.
But looks like it does not efficiently make things worse, you are right.

2. the solution is not easy, so, lets start with this one

Khatskevich (approved):

PR #260, opened 2024-11-14
Just download files from a single node in backup

joelynch:
Usually you don't want to download files from all nodes in backup - for this you can just run the tool multiple times anyway.


aris-aiven (approved):
lgtm

PR #261, opened 2024-11-15
Pause KeeperMap tables using read-only setting

aris-aiven:
Emitting a "table is read-only" is more precise
and is already handled by Kafka Sink Connector
for ClickHouse.

### DIFF
--- a/astacus/coordinator/plugins/clickhouse/steps.py
+++ b/astacus/coordinator/plugins/clickhouse/steps.py
@@ -181,36 +181,33 @@ class RetrieveUserDefinedFunctionsStep(Step[Sequence[UserDefinedFunction]]):
 
 
 @dataclasses.dataclass
-class KeeperMapTablesReadOnlyStep(Step[None]):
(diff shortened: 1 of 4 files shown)


astacus/coordinator/plugins/clickhouse/steps.py
```diff
@@ -257,7 +254,7 @@ async def run_step(self, cluster: Cluster, context: StepsContext) -> Sequence[Ke
 
     async def handle_step_failure(self, cluster: Cluster, context: StepsContext) -> None:
         try:
-            await KeeperMapTablesReadOnlyStep(clients=self.clients, allow_writes=True).run_step(cluster, context)
+            await KeeperMapTablesReadOnlyStep(self.clients).run_step(cluster, context)
```
tilman-aiven, line 257: for visibility: this is rephrased from `allow_writes=True` to `_is_read_only=False`

tilman-aiven (approved):
LGTM

PR #262, opened 2024-11-27
Only run lightweight sync between replicas during restore [DDB-1415]

aris-aiven:
Lightweight replica sync skips parts merges, part mutations, metadata alter, part cloning from other shards, and pinned part UUID sync. This should be good enough for the cluster to be in a working state, leaving the remaining work to be done asynchronously.


bogdanp05 (approved):

Khatskevich:
Why was it necessary to revert 8ac07a4ebc19f3d5a9bb97b3963e696a1c523328 ?

aris-aiven:
@Khatskevich it broke integration tests since the setup is trying to connect to Azure over an empty bucket endpoint and times out after 30s

joelynch (approved):

PR #263, opened 2024-12-05
clickhouse: allow restoring experimental MergeTree settings

aris-aiven:
The two settings were introduced between v23.8 and v24.3. They are needed when restoring by clusters that have defined them at the table level.


astacus/coordinator/plugins/clickhouse/steps.py
```diff
@@ -685,6 +685,8 @@ def _create_dbs(client: ClickHouseClient) -> Iterator[Awaitable[None]]:
             b"SET allow_hyperscan=true",
             b"SET allow_simdjson=true",
             b"SET allow_deprecated_syntax_for_merge_tree=true",
+            b"SET allow_experimental_block_number_column=true",
+            b"SET allow_experimental_replacing_merge_with_cleanup=true",
```
Khatskevich, line 689: I do not think this helps. Those are merge tree settings and you are setting them as query level settings. Those settings do not exist and should cause UNKNOWN_SETTING error.

But on the other hand, table settings should have been preserved and should not have caused any error... I do not yet understand it...

Khatskevich: main conversation here https://aiven-io.slack.com/archives/C052RDBD1TP/p1733428433123549

aris-aiven: True, these settings should have been set at the table level.

